PR #4, opened 2021-05-04
InputStreamAddon, DVDDemuxClient: assume progressive streams

kszaq:
CE's Kodi sets these to true by default, ffmpeg demuxer sets them accordingly but addons cannot do that. Most of streams from addons are progressive, set these to progressive by default.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxClient.cpp
+++ b/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemuxClient.cpp
@@ -465,6 +465,8 @@ void CDVDDemuxClient::SetStreamProps(CDemuxStream *stream, std::map<int, std::sh
       streamVideo->fAspect = source->fAspect;
       streamVideo->iFpsScale = source->iFpsScale;
       streamVideo->iFpsRate = source->iFpsRate;
+      streamVideo->bUnknownIP = false;
+      streamVideo->bInterlaced = false;
     }
     streamVideo->iBitRate = source->iBitRate;
     if (source->ExtraSize > 0 && source->ExtraData)
--- a/xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp
+++ b/xbmc/cores/VideoPlayer/DVDInputStreams/InputStreamAddon.cpp
@@ -419,6 +419,8 @@ KODI_HANDLE CInputStreamAddon::cb_get_stream_transfer(KODI_HANDLE handle,
     videoStream->iHeight = stream->m_Height;
     videoStream->fAspect = stream->m_Aspect;
     videoStream->iBitRate = stream->m_BitRate;
+    videoStream->bUnknownIP = false;
+    videoStream->bInterlaced = false;
     videoStream->profile = ConvertVideoCodecProfile(stream->m_codecProfile);
 
     /*! Added on API version 2.0.8 */


arthur-liberman:
Is this only relevant to Kodi in CE? And this doesn't occur on other platforms?
If so, maybe we should correct the original issue? Or is it not possible?

kszaq:
It's only relevant to Kodi in CE because of this modification: https://github.com/CoreELEC/xbmc/commit/1901eed81b063a49af2dcca5f46ec6e0eb1dde1c#diff-37d355a9f97dbc9001d30ecfdee2e761582e165c59790717b6e520c36cbc7f0aR131-R132

Upstream Kodi doesn't have these flags.

I guess @afl1's idea was to assume all streams are interlaced until proven progressive, due to most Live TV being interlaced.

Portisch:
Should it not be ok when ffmpeg does analyse the streams to general set the flags to `false`?
https://github.com/CoreELEC/xbmc/blob/aml-4.9-19.1/xbmc/cores/VideoPlayer/DVDDemuxers/DVDDemux.h#L131-L132

afl is absent since his last holiday trip and not seen anymore :disappointed:

kszaq:
When the stream is provided by Inputstream (Adaptive), it is not processed by FFMpeg, all demuxing is done in the addon.

Without this PR, SW-decoded streams are also affected: when you play anything from Netflix, it is reported as interlaced. I still need to check whether this enables SW-deinterlacing. If yes, this PR may have a positive impact on NF performance.

Edit: Checked, seems to have no impact, for streams decoded by Widevine this change would probably be cosmetic.

kszaq:
I now read your question again and noticed you asked about a different thing...

If it weren't for Live TV, I would opt for keeping them at `false` by default because like you said ffmpeg should set the `bInterlaced` flag to `true` if required. Also, if this is changed, the `bUnknownIP` flag can be removed because it is no longer used.

But since I do not have access to any interlaced content I decided not to modify the code from @afl1. I guess this modification was required because Amlogic's decoder may not good at detecting every kind of interlaced content.

If you want, I can upload the branch with these flags default on `false` today evening but if you had no bug reports about falsely detected interlaced content, I would suggest using the approach from this PR.

Portisch:
I checked a bit the src and would say a change of the default from `true` to `false` is a bad move.
Default CE Kodi does currently set interlaced on 25fps and 29.97fps content what aren't handled by ffmpeg demuxer.

Best solution would be the demuxer of the addon does report if the content is progressive or interlaced to make a correct handling. I am not sure what else will break when we set default progressive.

kszaq:
AFAIK there is no API from add-on to Kodi to report whether the content is interlaced or not, you would have to change the API. `bInterlaced` and `bUnknownIP` are CE-specific additions.

This PR touches only inputstream add-ons, I don't understand why would it break anything?

I would argue that this PR brings back upstream Kodi behaviour for content provided by inputstream add-ons - in my case it is "unbreaking" identification of 1080p25 content.

Portisch:
I am just afraid open other issues with other media content. What happens if 1080i50 is used or any other content what is 25fps interlaced? Will it be deinterlaced?

kszaq:
So far I have not seen any interlaced content provided by inputstream. 😉 

After looking at some issues with falsely deinterlacing progressive content, I believe that it's the decoder that decides whether to deinterlace the content or not. What we give it at start are only hints but during the playback it can switch from 25p to 50i: when tracking down the issue related to https://github.com/CoreELEC/media_modules-aml/pull/1, Kodi sent hints for 1080p25 but AMLCodec after a moment switched to 1080i50.

So to answer your last question: yes, I think that 1080i50 content will be deinterlaced.

All in all, I now start to think that this PR only changes what user sees in PlayerProcessInfo but does not change whether the decoder works in interlaced/deinterlaced mode.

Let me check one more thing related to this issue later this week.

Edit: I found mpd with 1080i50 content: http://rdmedia.bbc.co.uk/dash/ondemand/testcard/ Will check it later.

kszaq:
After checking:

HW-decoded mpd from InputStream: 1080i50 and 1080p25 are properly detected with or without this patch because this information is given by decoder (AMLCodec or FFMpeg).

SW-decoded Neflix video: without this patch the Player Debug Info shows `1280x720i @25.000` (not correct), with this patch it is `1280x720p @25.000` (correct).

Below the mpd stream to check 1080i50:
[bbc_1080i.strm.txt](https://github.com/CoreELEC/xbmc/files/6444029/bbc_1080i.strm.txt)

To sum up, this PR only changes Player Debug Info to show "p" instead of "i" for IA-decoded videos.

boot2k3:
I found that some other streams also have wrong p/i info in PlayerDebug.
It was a wrong place where it tried to recognize if video is interlaced or not.
@kszaq please check this patch on your stream if it solved your issue too.
link updated
https://github.com/boot2k3/xbmc-1/commit/beddd8b7e0d2c51e740ce35b9eb77b3d4247d710
Your sample shows correct info for me with this patch above.

kszaq:
@boot2k3 Your patch makes perfect sense. On the other hand I'm not sure if reading this value every video frame affects performance. Would it be possible to call it elsewhere, only when PlayerDebug is displayed?

And there's another thing: Netflix videos (and every other decoded by widevine). Without this PR PlayerDebug shows it's interlaced.

boot2k3:
@kszaq I made some changes in HW interlace detection for PlayerDebug. Now it detects it only 4 times on stream start or change.
I tested it on all HW streams, IPTV, YouTube and video files and looks like it works correct now.
https://github.com/CoreELEC/xbmc/pull/5/commits/5a6d4a60de0793092ea78640fb6784761368adc7
Also tested together with your PR, looks like all is correct.
Please can you rebase and FP the PR again?

kszaq:
@boot2k3 👍

PR #7, opened 2022-12-22
AMLCodec: GetPicture improvement

kszaq:
Recently, I've been using CE almost exclusively to play content from addons relying on Inputstream Adaptive. With some videos, I was getting occasional frame skips but with 4K content the skips were unbearable. Strangely, with PlayerDebug or debug log turned on, playback was buttery smooth. I tried a few different things and eventually noticed that the prefill value is lower than in my old Kodi 17 builds. Back then, after lots of testing with users, we found 0.5 to be the "sweet spot" for smooth playback. Bumping this value fixed my issues with stutter but I need to note that due to lack of many samples, I haven't run many more tests to see if anything was broken.

For your consideration. 🙂

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
@@ -2471,18 +2471,19 @@ DRAIN:
     if (errno != EAGAIN)
       CLog::Log(LOGERROR, "CAMLCodec::DequeueBuffer - VIDIOC_DQBUF failed: {}", strerror(errno));
 
-    std::chrono::milliseconds elapsed(std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now() - now).count());
 
-    if (elapsed < std::chrono::milliseconds(waitTime))
-      std::this_thread::sleep_for(std::chrono::milliseconds(waitTime) - elapsed);
+    if (m_drain)
+    {
+      std::chrono::milliseconds elapsed(std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now() - now).count());
 
-    timeout = elapsed >= std::chrono::milliseconds(300);
+      if (elapsed < std::chrono::milliseconds(waitTime))
+        std::this_thread::sleep_for(std::chrono::milliseconds(waitTime) - elapsed);
 
-    if (m_drain && !timeout)
-      goto DRAIN;
-
-    if (m_drain && timeout)
-      CLog::Log(LOGDEBUG, LOGAVTIMING, "CAMLCodec::DequeueBuffer timeout!");
+      if (elapsed < std::chrono::milliseconds(300))
+        goto DRAIN;
+      else
+        CLog::Log(LOGDEBUG, LOGAVTIMING, "CAMLCodec::DequeueBuffer timeout!");
+    }
 
     return -errno;
   }
@@ -2490,7 +2491,7 @@ DRAIN:
   if (m_drain)
   {
     int waited = std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::system_clock::now() - now).count();
-    CLog::Log(LOGDEBUG, LOGAVTIMING, "CAMLCodec::DequeueBuffer waited:{:.3f}ms", waited / 1000.0);
+    CLog::Log(LOGDEBUG, LOGAVTIMING, "CAMLCodec::DequeueBuffer waited:{:.3f}ms", waited);
   }
 
   m_last_pts = m_cur_pts;
@@ -2522,7 +2523,7 @@ float CAMLCodec::GetTimeSize()
   if (bs.free_len < (bs.data_len >> 1))
     return 7.0;
 
-  return (float)(m_frameSizes.size() * am_private->video_rate) / UNIT_FREQ;
+  return static_cast<float>(m_frameSizes.size() * am_private->video_rate) / UNIT_FREQ;
 }
 
 CDVDVideoCodec::VCReturn CAMLCodec::GetPicture(VideoPicture *pVideoPicture)
@@ -2533,8 +2534,6 @@ CDVDVideoCodec::VCReturn CAMLCodec::GetPicture(VideoPicture *pVideoPicture)
     return CDVDVideoCodec::VC_ERROR;
 
   float timesize(GetTimeSize());
-  if(!m_drain && timesize < 0.2f)
-    return CDVDVideoCodec::VC_BUFFER;
 
   if (DequeueBuffer() == 0)
   {
@@ -2561,7 +2560,7 @@ CDVDVideoCodec::VCReturn CAMLCodec::GetPicture(VideoPicture *pVideoPicture)
   }
   else if (m_drain)
     return CDVDVideoCodec::VC_EOF;
-  else if (timesize < 1.0f)
+  else if (timesize < 7.0f)
     return CDVDVideoCodec::VC_BUFFER;
 
   return CDVDVideoCodec::VC_NONE;


Portisch:
Thank you for your contribution. Let's take the PR first as patch in nightly before merge to have more user experience.

kszaq:
This still needs some investigation, the modification fixed playback of one title but another one keeps stuttering until PlayerDebug or debug info is being displayed.

Portisch:
I observed the same, so there must be somewhere else cause this "micro stutter".

arthur-liberman:
@kszaq Did you try to skip back 10-30 seconds and then see if the stuttering goes away? It's something I found to be an issue with some of my media.
Also, in my case, if I pull up the menu or debug output, it starts stuttering for me.

kszaq:
I've tried seeking forward and backward and it did not solve my issue. After some looking into the code and experimenting, I've come up with a new approach that solves stutter with all 4K content from my addon. PR updated, please check with your media. 🙂

Portisch:
I am sorry but I think the current status of the patch isn't good. 😞 It happened now two times the picture just stopped and audio continued playing. Stopping and start again the current stream resolve the issue. I use Dvbviewer PVR backend with TS streams from my SAT>IP system.

I never have observed such behavior so I am almost sure it's related to this change. I will revert the temporary added patch and see in more long time test if the issue is gone again.

Long time tests are done on A311D2, Amlogic-ne.

kszaq:
Thank you for testing and reporting. I suspect the freeze may occur in `DequeueBuffer`, let us check if we can simply add a condition that `timeline > 0.0f` before we try to call `DequeueBuffer`. I'll provide an updated patch later this week, if it still works for me.

I noticed similar issues when rw/ff live streams using inputstream.adaptive but they've been present since always, so I thought they were unrelated. I'll check if any modification to this patch can also fix this issue.

boot2k3:
I have possible similar issue, but with some video files playback on start or after pause, but not for streams. And this patch also doesn't solve it. And I also had picture freeze two times with this patch for one stream.
So if you have something to test then you can ping me too. I have one video on that I can repeat issue in 1 of 10-15 cases on start.
In case of issues I see bigger vq value in debug info than in case of normal playback for the same video file. Don't know if this info could help.

kszaq:
~It looks like lowering timesize requirement to `0.1f` solves stuttering in my case. I'll be looking if I can get anything to work around freezing in my ff/rw cases.~

kszaq:
I have added a check to not call `DequeueBuffer` if buffer is empty. Otherwise, even with `timesize < 0.1f` condition my test streams were stuttering after seek. If this version of the patch still causes issues for you, I'm a bit out of ideas.

boot2k3:
For my test video in case of issue I have vq~93-97%, aq=99%
When it starts normally, vq ~86-90%, aq=99%

After your current change in case of stutter issue I have vq ~97-99% and aq competes with vq now for 99% limit. 
In normal case(playing without stuttering) aq is stable 99%, vq ~86-90%

Portisch:
So the "trick" maybe will be to wait for some frames in buffer before starting dequeue? So a dequeue start, stop, start, stop,... can be avoided?

boot2k3:
Update: With this patch I got picture freeze on some static video stream part.
But no idea if this patch related or not. There is nothing in kodi log without enabled debug.
```[ 4111.553439@5]- 0: !!!wrong seq_info2 0x0 mb_width/mb_height (0x0/0x0)
[ 4111.553444@5]- 0: set parameters error, init_flag: 0
[ 4111.627462@0]- 0: !!!wrong seq_info2 0x0 mb_width/mb_height (0x0/0x0)
[ 4111.627466@0]- 0: set parameters error, init_flag: 0
[ 4111.627554@5]- 0: num_ref_frames change from 3 to 0
[ 4111.627564@5]- 0: reference list error 1 frame count 188662 to skip 0 reflist_error_count 1
[ 4111.628151@2]- 0: h264_reset_bufmgr frame count 188662 to skip 0
[ 4111.628151@2]- 
[ 4111.668496@2]- H264 sysinfo: 1920x1080 duration=3844, pts_outside=1
[ 4111.668502@2]- sync_outside=0, use_idr_framerate=0, is_used_v4l: 0
[ 4111.668511@2]- 0: !!!wrong seq_info2 0x0 mb_width/mb_height (0x0/0x0)
[ 4111.669822@0]- 0: !!!wrong seq_info2 0x0 mb_width/mb_height (0x0/0x0)
[ 4111.669827@0]- 0: set parameters error, init_flag: 0```

**Comment by Portisch** [comment]
https://discourse.coreelec.org/t/amlogic-ne-new-era/19354/176
Temporary patch is removed again.

**Comment by kszaq** [comment]
In my feeling, `GetPicture` should always display and return `VC_PICTURE` as soon as possible and there is some issue with `VIDIOC_DQBUF`, probably in some conditions when there's not enough data in the buffer or with still frames. https://github.com/CoreELEC/xbmc/commit/bc0c714e24816ba09cead78d9fd2474024b0168e has so far been the best working approach for me but I understand that it can cause issues with other devices/media sources. I'll have a look if there is any other way to solve stuttering in my streams, otherwise I'll simply keep the patch for a personal Kodi build.

**Comment by kszaq** [comment]
I was looking at the `AMLCodec` and I don't quite understand the idea behind `GetTimeSize`. Until Krypton, this function was returning amount of buffered data in seconds. Since Leia, it takes amount of queued data and divides it by fps - is `AddData` working frame-by-frame? If yes, then this approach should still give us length of buffered data in seconds...
However, with this approach to return percentage of buffer fill instead of "length" of buffered data (no other modifications proposed in this PR), the stuttering is gone:
```
diff --git a/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp b/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
index 214d9f68c4..397d4b7e33 100644
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
@@ -2522,7 +2522,7 @@ float CAMLCodec::GetTimeSize()
   if (bs.free_len < (bs.data_len >> 1))
     return 7.0;
 
-  return (float)(m_frameSizes.size() * am_private->video_rate) / UNIT_FREQ;
+  return (float)(bs.data_len) / (float)(bs.size);
 }
 
 CDVDVideoCodec::VCReturn CAMLCodec::GetPicture(VideoPicture *pVideoPicture)
```
This patch with rest of the code should probably keep the codec buffer filled >=20%, which so far seems enough to remove stuttering, at least with the media I tried.

**Comment by Portisch** [comment]
Not really, `AddData` add encoded frame packets (whatever format) to the hardware decoder. Kodi read back the decoded single frames. That's also a issue for keeping up the PTS correct about raw frame encoded packet and decoded single frames as it's not known before decode.

**Comment by kszaq** [comment]
So basically `GetTimeSize` is not really correct now as it's based on number of encoded data packets queued and not actual frames. This might be the reason why `GetPicture` does not always make good decision whether there is enough data buffered to dequeue and we get stuttering.

I'll be checking if this makes sense with the plugin I'm trying to fix playback for.

**Comment by Portisch** [comment]
Sounds reasonable. This code was every time I remember like this. So you update the PR that we can add it again temporary for nightly tests?

**Comment by kszaq** [comment]
Done, PR updated. While it works smoothly with my media, it requires a lot of testing with different sources and formats.

**Comment by boot2k3** [comment]
With this change it always stuttering on first tested video with Debug Info enabled.
And after seek picture freezed.

**Comment by kszaq** [comment]
@boot2k3 Thanks for testing. The calculation is wrong. I'll check again.

**Comment by kszaq** [comment]
@boot2k3 PR updated with a fix in calculations. Previously, I checked with Jellyfish test video and confirmed your observations. Now it works fine.

**Comment by kszaq** [comment]
Please don't use it yet, I checked with a very low bitrate video, there's an issue with playback.

**Comment by boot2k3** [comment]
Yep, I also checked it on my test video and it has a lot of skip frames in the beginning and av is about 0% on start.
In case of stuttering with enabled debug info it has av value ~90%

**Comment by kszaq** [comment]
Back to old approach, because looking at the size of buffered data may only be good with moderate to high bitrate videos but does not work with low bitrate. 

Differences from https://github.com/CoreELEC/xbmc/commit/bc0c714e24816ba09cead78d9fd2474024b0168e which had the freeze issue:
* always call `GetTimeSize()` in `GetPicture()` to dequeue elements from `m_frameSizes` - may prevent a freeze
* do not wait 5ms in unsuccessful `DequeueBuffer()` if not draining - `VC_BUFFER` is returned sooner
* allow to buffer more than ca. 1s of data

**Comment by boot2k3** [comment]
On start now there is 10-15% of vq, after some time it increases to ~60-75% and plays normal from first view.
But after 10-15 pauses it goes unexpectedly to 96-99% and starts stuttering with debug info enabled.. and after next pauses it is still on same level 97-99%
So looks like stuttering always occur if buffer is on it limit. And no idea why does it unexpectedly increases after pause or on start sometimes..

**Comment by Portisch** [comment]
There was a change some time ago about not flood the buffer when pause.
Maybe this need to be adjusted to match the new design?
https://github.com/CoreELEC/xbmc/commit/47bd5fed39e3a20912ea543e9b097f9e23267ce6

**Comment by boot2k3** [comment]
And one opposite case with this patch.. it stutters sometimes for 4k stream and doesn't stutter with debug info enabled.
aq~23%, vq\~33%, so not related to buffer problem, but related with this changes..

**Comment by kszaq** [comment]
@boot2k3 This is exactly an issue that I'm trying to solve and it's better with this patch. But if it's not solved for you then real cause is somewhere else and I'm out of ideas.

**Comment by kszaq** [comment]
@boot2k3 I tried pausing-playing a 4K video a few times and the stutter "returned". Modifying `GetPicture` looks to be a way to mask the issue if playback is started and not paused but then it's back, so the real cause is not fixed. It's probably something else completely.

**Comment by boot2k3** [comment]
No idea why it happens only on start and after pause.
Portisch flood protection part looks good. May be some issue happens when it receives new packets and couldn't sync av part before buffer is full?

**Comment by Portisch** [comment]
Another issue might be that the return value of the vbuf isn't checked:
https://github.com/CoreELEC/xbmc/blob/aml-4.9-20.0/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp#L2511

It's possible the buffer request return **-EINVAL**:
https://github.com/CoreELEC/media_modules-aml/blob/amlogic-4.9.269/drivers/stream_input/amports/amstream.c#L2554

Also there do exist `VFRAME_BLOCK_MIN_LEVEL` and `VFRAME_BLOCK_MAX_LEVEL`:
https://github.com/CoreELEC/media_modules-aml/blob/amlogic-4.9.269/drivers/frame_provider/decoder/utils/vdec_input.c#L699-L719
So this kind of buffer check is maybe needed as well on kodi side to know if it's needed to feed new am_pkt by AddData or not.

Does there a sample exist with 50-60Hz as this is the only frame rate I can test here local.

**Comment by kszaq** [comment]
Unforunately, I don't have any sample to test as I'm seeing this issue only with DRM-protected (albeit hw-decoded) streams that I can't record.

FYI, I'm not testing 2 modifications: 
1. [AMLCodec simplification](https://github.com/kszaq/xbmc-ce/commit/873db07b9b3dbe17c8e58eb335312474a4cef94e) to simply allow buffer the data if there's free space available. It seemed to me that a duplicate check in `GetPicture`/`GetTimesize` is an overkill.
2. [RenderManager revert to "stock" Kodi](https://github.com/kszaq/xbmc-ce/commit/d07f171f79564dfcea64a3897048f93b3d2f760b) to see if it helps after VideoCodecAmlogic is a bit closer to how other decoders work in `AddData` and `GetPicture`.

**Comment by kszaq** [comment]
FWIW the stutter is present only when using 0.5s "Delay after refresh rate change". With no or 2s delay it's gone. It's not gone completely, occasionally it's back. There's no clear indication in the debug log why this is happening. 

Any patch that I proposed in this PR seemed to simply decrease/increase a delay before decoder started outputting frames, so they all simply masked the issue but did not provide a real fix.

With a lower resolution video and 0.2s delay, I can see a frame drop at the beginning but afterwards it's OK. I am wondering if with a higher resolution/complexity, video decoder is trying to catch up but can't process the frames as quickly as required. 

Since I now have a workaround, this became low priority but I'll let you know if I find a reliable way to reproduce the issue and/or a possible solution.

**Comment by kszaq** [comment]
@Portisch How is frame skip/drop handled after https://github.com/CoreELEC/linux-amlogic/commit/53c565ee219af353944de88be46618da799c70ff ?

**Comment by Portisch** [comment]
Long time ago...
I searched a bit and it solved skipped frames on pulling them from kernel to Kodi.
Now on every call it pop all done packets from queue to ready queue, ready to be picked up from Kodi.

**Comment by boot2k3** [comment]
May be on start/resume some of them should be skipped to not overflow aq or vq level? But no idea about the reason when to do it..

**Comment by kszaq** [comment]
@Portisch Do you remember a reason why this was changed? It feels like because of this change, a request to drop from Kodi does nothing. With revert and  modification from to use `amlvideo_vf_put` it seems to be working:
```
diff --git a/drivers/amlogic/media/video_processor/video_dev/amlvideo.c b/drivers/amlogic/media/video_processor/video_dev/amlvideo.c
index 741b48cd4153..6c9411e01693 100644
--- a/drivers/amlogic/media/video_processor/video_dev/amlvideo.c
+++ b/drivers/amlogic/media/video_processor/video_dev/amlvideo.c
@@ -554,16 +554,29 @@ static int vidioc_qbuf(struct file *file, void *priv, struct v4l2_buffer *p)
 	while ((vf = vfq_peek(&dev->q_omx)))
 	{
 		index = (u32)vf->pts_us64;
-		vfq_push(&dev->q_ready, vfq_pop(&dev->q_omx));
-		ATRACE_COUNTER(dev->v4l2_dev.name, vfq_level(&dev->q_omx));
-		ATRACE_COUNTER(dev->v4l2_dev.name, vfq_level(&dev->q_ready));
-		vf_notify_receiver(
-				dev->vf_provider_name,
-				VFRAME_EVENT_PROVIDER_VFRAME_READY,
-				NULL);
-
-		if (p->index == index)
-			break;
+		if (p->index > index)
+		{
+			amlvideo_vf_put(vfq_pop(&dev->q_omx), dev);
+			ATRACE_COUNTER(dev->v4l2_dev.name, vfq_level(&dev->q_omx));
+			printk("vidioc_qbuf skip: index:%u:%u\n", p->index, index);
+			continue;
+		}
+		else if (p->index == index)
+		{
+			vf = vfq_pop(&dev->q_omx);
+			ATRACE_COUNTER(dev->v4l2_dev.name, vfq_level(&dev->q_omx));
+			if (p->flags & V4L2_BUF_FLAG_DONE)
+				amlvideo_vf_put(vf, dev);
+			else {
+				vfq_push(&dev->q_ready, vf);
+				ATRACE_COUNTER(dev->v4l2_dev.name, vfq_level(&dev->q_ready));
+				vf_notify_receiver(
+					dev->vf_provider_name,
+					VFRAME_EVENT_PROVIDER_VFRAME_READY,
+					NULL);
+			}
+		}
+		break;
 	}
 	return 0;
 }
```
Naturally, my testing is very limited so I probably miss something.

**Comment by Portisch** [comment]
Reason:
```text
Jan 01 00:36:34.976747 CoreELEC kernel: vidioc_qbuf skip: index:1841:1840
Jan 01 00:36:34.996739 CoreELEC kernel: vidioc_qbuf skip: index:1844:1843
Jan 01 00:36:35.016754 CoreELEC kernel: vidioc_qbuf skip: index:1846:1845
Jan 01 00:36:35.036760 CoreELEC kernel: vidioc_qbuf skip: index:1848:1847
```
`when 1844 will be released before 1843 it will be a message becuase of if (p-&gt;index &gt; index) in amlvideo.c`

From searching the history:
`best way to test qbuf skips is with the 60fps testvideo on the N2`

And after start debugging with this patch:
`now only the dropped frame get lost, not 2 or more like before`

And
`running smooth with 1.99` pointing to here: https://github.com/CoreELEC/xbmc/blob/aml-4.9-20.0/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.cpp#L1205

But I see there is 0.98 and not 1.99. This value was set by AFL:
https://github.com/CoreELEC/CoreELEC/blob/coreelec-9.0/packages/mediacenter/kodi/patches/amlogic-3.14/kodi-le-030-lateframes.patch

So maybe this is a missing part? AFL is lost and it was from times adding just patches without commit, so no idea about the sense to increase the value. But the value was never changed in 19.x so I have no idea anymore why we did not use this value anymore.

PR #9, opened 2023-02-02
AMLCodec: signal VC_FLUSHED if buffer gets empty

kszaq:
This is a follow-up to https://github.com/CoreELEC/xbmc/pull/7#issuecomment-1410620011

Normally, VideoPlayer should feed data to decoder to keep TimeSize between 0.2 and 1.0. If for some reason decoder is stuck/reset, buffer can starve and this can lead to stutter. In this case, signal VC_FLUSHED to allow VideoPlayer restart decoding and rendering.

This patch fixes "initial stutter" I can see when "Delay after refresh rate change" is at the point do make decoder derp.

In my case, right after player is unpaused, `bs.data_len` is unexpectedly down to `0` and decoding/rendering gets out of sync:
```
debug <general>: SetCodecControl 4000000->0
debug <general>: CAMLCodec::AddData: dl:570710 sum:570710 sz:86528 dts:0.720 pts:0.720
debug <general>: CAMLCodec::GetTimeSize: len:7 dl:570710 fs:570710 front:93467
debug <general>: CAMLCodec::DequeueBuffer: pts:0.440  idx:218
debug <general>: CAMLCodec::GetPicture: index: 218, pts: 0.440, dur:0.000ms ar:1.80
debug <general>: CAMLCodec::ReleaseFrame idx:217, drop:1
debug <general>: CAMLCodec::PollFrame elapsed:15.004ms
debug <general>: PrepareNextRender: frameOnScreen: -0.012 renderPts: 0.068 nextFramePts: 0.120 -> diff: -0.052  render: 0 forceNext: 
debug <general>: CAMLCodec::PollFrame elapsed:9.216ms
debug <general>: PrepareNextRender: frameOnScreen: 0.011 renderPts: 0.091 nextFramePts: 0.120 -> diff: -0.029  render: 0 forceNext: 0
debug <general>: CVideoPlayer::ProcessVideoData size:86057 dts:0.760 pts:0.760 dur:40.000ms, clock:0.014 level:0
debug <general>: CAMLCodec::PollFrame elapsed:7.440ms
debug <general>: PrepareNextRender: frameOnScreen: 0.027 renderPts: 0.107 nextFramePts: 0.120 -> diff: -0.013  render: 0 forceNext: 0
debug <general>: CAMLCodec::ReleaseFrame idx:210, drop:0
debug <general>: CAMLCodec::SetVideoRect:display(0,0,3840,2160)
debug <general>: CAMLCodec::SetVideoRect:gui(0,0,1920,1080)
debug <general>: CAMLCodec::SetVideoRect:m_dst_rect(0,4,1920,1072)
debug <general>: CAMLCodec::SetVideoRect:dst_rect(0,8,3840,2144)
debug <general>: CAMLCodec::SetVideoRect:m_guiStereoMode(0)
debug <general>: CAMLCodec::SetVideoRect:m_guiStereoView(0)
debug <general>: CAMLCodec::PollFrame elapsed:10.152ms
debug <general>: PrepareNextRender: frameOnScreen: 0.044 renderPts: 0.124 nextFramePts: 0.160 -> diff: -0.036  render: 0 forceNext: 0
debug <general>: CVideoPlayerVideo::OutputPicture - ttd:461ms pts:0.440 Clock:-0.022 Level:6
debug <general>: CAMLCodec::GetTimeSize: len:7 dl:0 fs:570710 front:93467
debug <general>: CAMLCodec::GetTimeSize: len:6 dl:0 fs:477243 front:85638
debug <general>: CAMLCodec::GetTimeSize: len:5 dl:0 fs:391605 front:59298
debug <general>: CAMLCodec::GetTimeSize: len:4 dl:0 fs:332307 front:63175
debug <general>: CAMLCodec::GetTimeSize: len:3 dl:0 fs:269132 front:92144
debug <general>: CAMLCodec::GetTimeSize: len:2 dl:0 fs:176988 front:90460
debug <general>: CAMLCodec::GetTimeSize: len:1 dl:0 fs:86528 front:86528
debug <general>: CAMLCodec::GetTimeSize: len:0 dl:0 fs:0 front:959524708
```
`FLUSH` is a workaround but seems to be an effective one.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
@@ -2510,6 +2510,8 @@ float CAMLCodec::GetTimeSize()
   struct buf_status bs;
   m_dll->codec_get_vbuf_state(&am_private->vcodec, &bs);
 
+  bool frameSizesEmpty = m_frameSizes.empty();
+
   CLog::Log(LOGDEBUG, LOGVIDEO, "CAMLCodec::GetTimeSize: len:{:d} dl:{:d} fs:{:d} front:{:d}",
     m_frameSizes.size(), bs.data_len, m_frameSizeSum, m_frameSizes.front());
   while (m_frameSizeSum >  (unsigned int)bs.data_len)
@@ -2519,6 +2521,11 @@ float CAMLCodec::GetTimeSize()
     CLog::Log(LOGDEBUG, LOGVIDEO, "CAMLCodec::GetTimeSize: len:{:d} dl:{:d} fs:{:d} front:{:d}",
       m_frameSizes.size(), bs.data_len, m_frameSizeSum, m_frameSizes.front());
   }
+
+  // If the queue was emptied, decoder starved
+  if (m_frameSizes.empty() && !frameSizesEmpty)
+    return -1.0f;
+
   if (bs.free_len < (bs.data_len >> 1))
     return 7.0;
 
@@ -2533,6 +2540,11 @@ CDVDVideoCodec::VCReturn CAMLCodec::GetPicture(VideoPicture *pVideoPicture)
     return CDVDVideoCodec::VC_ERROR;
 
   float timesize(GetTimeSize());
+
+  // If decoder starved, reset to avoid stuttering
+  if (!m_drain && timesize < 0.0f)
+    Reset();
+
   if(!m_drain && timesize < 0.2f)
     return CDVDVideoCodec::VC_BUFFER;
 


boot2k3:
Looks promising.
Will do more tests, but after 25 tests with video start I couldn't repeat issue on my test video.
But after pause/resume I don't see stuttering, but sometimes still see a/v unsync with higher vq value.

kszaq:
Without this patch, does desync also happen? In my feeling yes, so it's not related... Regarding desync, I am now checking if reintroducing frame drop in kernel (without frame skip) might help.

boot2k3:
Yes, it was unsync + stuttering before. With this patch stuttering went off for this case.
One issue found with this patch. Sometimes live stream stuck on start with pc:8 state in debuginfo
[kodi.log](https://github.com/CoreELEC/xbmc/files/10608489/kodi.log)

kszaq:
@boot2k3 Can you disable ffmpeg logging and enable video debug logging?

Portisch:
For me with using DVBViewer PVR and SAT>IP system it's an improvement. But there are other users reporting issues:
https://discourse.coreelec.org/t/amlogic-ne-new-era/19354/321

boot2k3:
@kszaq, log is in attached file with video debug enabled
[kodi_video.log](https://github.com/CoreELEC/xbmc/files/10609676/kodi_video.log)
Looks like sometimes it's in the `video decoder was flushed` loop

kszaq:
Thanks, I can see where the issue is and how to fix it. AFK now, I'll provide a new patch next week.

kszaq:
@boot2k3 Can you try adding an additional condition to `if` in the patch: `&& m_cur_pts != DVD_NOPTS_VALUE`? This value is set when first frame is dequeued from decoder, so `VC_FLUSHED` would not be signalled if decoder is waiting for a keyframe.

boot2k3:
@kszaq , with this change it has the same stuttering issue for Live TV stream

kszaq:
@boot2k3 Can you please try with the updated patch?

BTW regarding the desync, I'm checking if reintroducing frame drop kernel-side may help: http://ix.io/4ng4 
As @Portisch advised, frame skip should still be handled by Kodi with this patch.

boot2k3:
With updated patch still video stuck on live TV start sometimes.
http://ix.io/4ngy
Desync patch I will test too soon

desync patch failed to apply, applied only after some line changes.

kszaq:
> With updated patch still video stuck on live TV start sometimes. http://ix.io/4ngy 

Thank you for testing. Looking at the log, the buffer is starved 7 times - does this video start without the patch, is it working normally?

> Desync patch I will test too soon
> 
> desync patch failed to apply, applied only after some line changes.

The patch is for 4.9 kernel, because I only have a S905X3 device.

boot2k3:
I tested now with only desync patch on 5.4 kernel and I couldn't repeat stuttering on my test file today. And also don't see this unsync issue.
With higher vq value too. Need more testing.

> Thank you for testing. Looking at the log, the buffer is starved 7 times - does this video start without the patch, is it working normally?

Yes, it's live stream and it works without this PR patch and with desync patch too..

Portisch:
Can have this also affect?
https://github.com/CoreELEC/xbmc/blob/amlogic-3.14-18.9/xbmc/cores/VideoPlayer/VideoRenderers/RenderManager.cpp#L1146
It's not used in 4.9 or newer. So maybe it needs to be increased to `1.99` again?

kszaq:
@Portisch I haven't noticed a difference between 0.98 and 1.99.

kszaq:
> I tested now with only desync patch on 5.4 kernel and I couldn't repeat stuttering on my test file today. And also don't see this unsync issue. With higher vq value too. Need more testing.
> 
> > Thank you for testing. Looking at the log, the buffer is starved 7 times - does this video start without the patch, is it working normally?
> 
> Yes, it's live stream and it works without this PR patch and with desync patch too..

Unfortunately, without a Kodi patch my test video stutters even with a kernel patch applied.

It would be great if you could post a debug log from playing your live TV stream without the Kodi patch for comparison.

boot2k3:
@kszaq, this is a log without this PR and with desync patch on live stream
http://ix.io/4ngL

I can share a stream for test if you give me your IP address, to add it in allow list.
May be in PM in discourse forum.

kszaq:
Thank you. So basically, this PR is wrong because it looks like decoder can still work fine even if input buffer is empty. Perhaps `FLUSHED` needs to be signalled if both input and output buffers are empty, but I don't know yet how to check it.

boot2k3:
> Unfortunately, without a Kodi patch my test video stutters even with a kernel patch applied.

I agree, without this patch it's still stutter sometimes.
Need some other approach, but already there is big progress.

kszaq:
@boot2k3 If you can, one last force-pushed patch to test with your live stream, please. 🙂 

If it doesn't work, I promise I'll give up. 😄

boot2k3:
Still freezes on live stream start
http://ix.io/4nka
Please don't give up. I'm ready to test your builds untill it will be fixed with some pause on 08-15 February vacation)

May be just some parameters and not all decoder should be reseted?

kszaq:
End of the day, the last approach I can propose, which should be the least invasive:
https://github.com/kszaq/xbmc-ce/commit/c72a96ae62750dd95085cba95a7c88d6c69c4840
It seems that if we observe a few frames removed from decoder input buffer at once, decoder consumed them and there should be an output picture we can dequeue. If instead we ask VP for more data, this can lead to stutter, because we wait for more data and decoded picture is displayed too late. At least this helps with my stuttering videos and should not break other cases.

There are also 2 really minor patches I can propose:
https://github.com/kszaq/xbmc-ce/commit/151a0171adf5d81dace2a7be6cc06f6796515221
https://github.com/kszaq/xbmc-ce/commit/aba3d32c8078bce7312501595980d5f9f32b7bfe

Portisch:
I currently hunting "another" frame display issue. Like when timebar is visible it stuck like every 0.5-1s.
I spotted in kodi log `PollFrame` is different when this happen:
Like this example:
```text
	Line 12275: 2023-02-13 09:34:28.087 T:873     debug <general>: CAMLCodec::PollFrame elapsed:18.466ms
	Line 12287: 2023-02-13 09:34:28.107 T:873     debug <general>: CAMLCodec::PollFrame elapsed:18.436ms
	Line 12300: 2023-02-13 09:34:28.127 T:873     debug <general>: CAMLCodec::PollFrame elapsed:16.379ms
	Line 12312: 2023-02-13 09:34:28.149 T:873     debug <general>: CAMLCodec::PollFrame elapsed:0.011ms, < 1000/60ms
	Line 12325: 2023-02-13 09:34:28.168 T:873     debug <general>: CAMLCodec::PollFrame elapsed:0.009ms, < 1000/60ms
	Line 12334: 2023-02-13 09:34:28.188 T:873     debug <general>: CAMLCodec::PollFrame elapsed:0.010ms, < 1000/60ms
	Line 12347: 2023-02-13 09:34:28.207 T:873     debug <general>: CAMLCodec::PollFrame elapsed:16.239ms
```
So when it works the poll take ~18ms. When the picture "stuck" it take almost nothing.
I don't know the reason for it and where the poll actually goes into. I think in amports of media modules, but I am not sure. Are there no packets in decoder waiting for in the poll to be decoded? Or are frames already ready when the poll immediately returns? Magic?

Libamcodec open the decoder file like `/dev/amstream_hevc_frame` and return the handle to AMLCodec. And this handle is polled so it must go into like [here](https://github.com/CoreELEC/media_modules-aml/blob/amlogic-5.4.125/drivers/stream_input/amports/amstream.c#L403). But `vframe_fops` do not include a `poll`, so where it do actually goes into?

EDIT:
The poll goes here:
https://github.com/CoreELEC/linux-amlogic/blob/amlogic-5.4.125/drivers/amlogic/media/video_sink/video.c#L11674

For tests I applied your latest patch but this does not show any improvement about this issue.
But I am sure it's a different reason but maybe the issues are linked together somehow.

kszaq:
Have you tried to revert the `PollFrame` part of https://github.com/CoreELEC/xbmc/commit/fdf2017648782f1d0e45631d606555ae931a2756 ?

Portisch:
Not yet, I found out it's about the GUI rendered frames. Every time a frame is rendered it skip one frame output until it's to much and queued frames get dropped. I need to look more in detail how this works, maybe the frame from hardware decoder need to be replaced by the frame from GUI renderer? Now it looks like the frame is squeezed between the hw frames so pts get more and more late until they are dropped.

kszaq:
> End of the day, the last approach I can propose, which should be the least invasive: [kszaq@c72a96a](https://github.com/kszaq/xbmc-ce/commit/c72a96ae62750dd95085cba95a7c88d6c69c4840) It seems that if we observe a few frames removed from decoder input buffer at once, decoder consumed them and there should be an output picture we can dequeue. If instead we ask VP for more data, this can lead to stutter, because we wait for more data and decoded picture is displayed too late. At least this helps with my stuttering videos and should not break other cases.
> 
Before you start testing: this is not working. 🤣 Seems like yet another way to mask the issue rather than solve it. Yesterday I played 1080p H.264 on-demand stream (not live) and it stuttered heavily after pause-resume.

Portisch:
So I should remove this again?
https://github.com/CoreELEC/CoreELEC/commit/d1957138dc1a4e99aed30482bcb5507bf8c64104

kszaq:
It does not hurt and help in a limited number of situations but does not solve the root cause.

Portisch:
I have seen indeed some "stuck" on live TV so I removed the patches for now.
How I can reproduce or trigger the issue or what I should I observe?
I did a lot with frame handling last days so I am again "in flow" how it's handled and may be able to help.

Portisch:
On first look in mmm I saw the `buf_status` is get by `codec_get_vbuf_state`. This ends up in mmm [vdec_input_get_status](https://github.com/CoreELEC/media_modules-aml/blob/amlogic-5.4.125/drivers/frame_provider/decoder/utils/vdec_input.c#L589) and finally at [vdec_input_level](https://github.com/CoreELEC/media_modules-aml/blob/amlogic-5.4.125/drivers/frame_provider/decoder/utils/vdec_input.c#L625). And there the "circular buffer" is handled as one way buffer. Both `total_wr_count` and `total_rd_count` are `u64`. So when `total_wr_count` have an overflow the result of the function is just garbage. There is just no handling if read position is in front of writing position. This is may a reason of disturb kodi flow and cause dropping frames. Maybe...

So this may help if the buffer read position becomes higher than the write position (wr position overflow): (UNTESTED!)
```c
diff --git a/drivers/frame_provider/decoder/utils/vdec_input.c b/drivers/frame_provider/decoder/utils/vdec_input.c
index 6fec0ad9..d7616956 100644
--- a/drivers/frame_provider/decoder/utils/vdec_input.c
+++ b/drivers/frame_provider/decoder/utils/vdec_input.c
@@ -622,7 +622,10 @@ static inline void vdec_input_del_block_locked(struct vdec_input_s *input,
 
 int vdec_input_level(struct vdec_input_s *input)
 {
-       return input->total_wr_count - input->total_rd_count;
+       if (input->total_wr_count < input->total_rd_count)
+               return (input->size - input->total_rd_count) + input->total_wr_count;
+       else
+               return input->total_wr_count - input->total_rd_count;
 }
 EXPORT_SYMBOL(vdec_input_level);
 
@@ -974,7 +977,10 @@ int vdec_input_add_chunk(struct vdec_input_s *input, const char *buf,
        }
        if (chunk->size > input->frame_max_size)
                input->frame_max_size = chunk->size;
-       input->total_wr_count += count;
+       if (input->total_wr_count + count > input->size)
+               input->total_wr_count = count - (input->size - input->total_wr_count);
+       else
+               input->total_wr_count += count;
        vdec_input_unlock(input, flags);
 #if 0
        if (add_count == 2)
@@ -1102,7 +1108,10 @@ void vdec_input_release_chunk(struct vdec_input_s *input,
        block->data_size -= chunk->size;
        block->chunk_count--;
        input->data_size -= chunk->size;
-       input->total_rd_count += chunk->size;
+       if (input->total_rd_count + chunk->size > input->size)
+               input->total_rd_count = chunk->size - (input->size - input->total_rd_count);
+       else
+               input->total_rd_count += chunk->size;
        if (block->is_out_buf) {
                list_move_tail(&block->list,
                        &input->vframe_block_free_list);
```

boot2k3:
@Portisch , just tested with your patch and see some strange changes.
In case of stuttering after pause it stutter constantly now. But if I enable DebugInfo or some GUI elements are active, then there is no any stuttering anymore.

kszaq:
> I have seen indeed some "stuck" on live TV so I removed the patches for now. How I can reproduce or trigger the issue or what I should I observe? I did a lot with frame handling last days so I am again "in flow" how it's handled and may be able to help.

Unfortunately, I have found no way to reproduce the issue other than playing a video from a plugin with geolocked content, so that won't help you much. I'll try to dig in my test videos archive to see if it's possible to reproduce the issue with any of them.

boot2k3:
> In case of stuttering after pause it stutter constantly now. But if I enable DebugInfo or some GUI elements are active, then there is no any stuttering anymore.

Looks like this stuttering was related to `m_guiAlgorithmDirtyRegions = 0;` change.
Will do more tests, as now it reverted.

Portisch:
@kszaq can you test these "dirty" changes? It's not final version but it sync the buffer handling between Kodi and vdec. So the "frame sizes" what are chunk sizes do match. Also the chunk data on header only pkt were missing at all.
Kodi:
```diff
diff --git a/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp b/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
index b46ce5035a..aa1476e9ce 100644
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
@@ -44,6 +44,11 @@
 #include "aom_integer.h"
 #include "obu_util.h"
 
+#define L1_CACHE_SHIFT         (6)
+#define L1_CACHE_BYTES         (1 << L1_CACHE_SHIFT)
+#define MIN_FRAME_PADDING_SIZE ((int)(L1_CACHE_BYTES))
+#define PAGE_SIZE              (4096)
+
 namespace
 {
 
@@ -58,6 +63,9 @@ extern "C" {
 
 CEvent g_aml_sync_event;
 
+std::deque<uint32_t> m_frameSizes;
+std::uint32_t m_frameSizeSum;
+
 class PosixFile
 {
 public:
@@ -605,6 +613,7 @@ int check_in_pts(am_private_t *para, am_packet_t *pkt)
 static int write_header(am_private_t *para, am_packet_t *pkt)
 {
     int write_bytes = 0, len = 0;
+    int need_padding_size = MIN_FRAME_PADDING_SIZE;
 
     if (pkt->hdr && pkt->hdr->size > 0) {
         if ((NULL == pkt->codec) || (NULL == pkt->hdr->data)) {
@@ -621,6 +630,16 @@ static int write_header(am_private_t *para, am_packet_t *pkt)
         }
         while (1) {
             write_bytes = para->m_dll->codec_write(pkt->codec, pkt->hdr->data + len, pkt->hdr->size - len);
+            if (write_bytes < PAGE_SIZE) {
+              need_padding_size += PAGE_SIZE - ((write_bytes + need_padding_size) & (PAGE_SIZE - 1));
+            } else {
+              /*to 64 bytes aligned;*/
+              if (write_bytes & 0x3f)
+                need_padding_size += 64 - (write_bytes & 0x3f);
+            }
+            m_frameSizes.push_back(write_bytes + need_padding_size);
+            m_frameSizeSum += (write_bytes + need_padding_size);
+
             if (write_bytes < 0 || write_bytes > (pkt->hdr->size - len)) {
                 if (-errno != AVERROR(EAGAIN)) {
                     CLog::Log(LOGDEBUG, "ERROR:write header failed!");
@@ -653,6 +672,8 @@ int write_av_packet(am_private_t *para, am_packet_t *pkt)
     int write_bytes = 0, len = 0, ret;
     unsigned char *buf;
     int size;
+    int need_padding_size = MIN_FRAME_PADDING_SIZE;
+
 
     // do we need to check in pts or write the header ?
     if (pkt->newflag) {
@@ -679,6 +700,16 @@ int write_av_packet(am_private_t *para, am_packet_t *pkt)
 
     while (size > 0 && pkt->isvalid) {
         write_bytes = para->m_dll->codec_write(pkt->codec, buf, size);
+        if (write_bytes < PAGE_SIZE) {
+          need_padding_size += PAGE_SIZE - ((write_bytes + need_padding_size) & (PAGE_SIZE - 1));
+        } else {
+          /*to 64 bytes aligned;*/
+          if (write_bytes & 0x3f)
+            need_padding_size += 64 - (write_bytes & 0x3f);
+        }
+        m_frameSizes.push_back(write_bytes + need_padding_size);
+        m_frameSizeSum += (write_bytes + need_padding_size);
+
         if (write_bytes < 0 || write_bytes > size) {
             CLog::Log(LOGDEBUG, "write codec data failed, write_bytes({:d}), errno({:d}), size({:d})", write_bytes, errno, size);
             if (-errno != AVERROR(EAGAIN)) {
@@ -1786,11 +1817,11 @@ CAMLCodec::CAMLCodec(CProcessInfo &processInfo)
   , m_last_pts(DVD_NOPTS_VALUE)
   , m_bufferIndex(-1)
   , m_state(0)
-  , m_frameSizeSum(0)
   , m_processInfo(processInfo)
 {
   am_private = new am_private_t;
   memset(am_private, 0, sizeof(am_private_t));
+  m_frameSizeSum = 0;
   m_dll = new DllLibAmCodec;
   if(!m_dll->Load())
     CLog::Log(LOGWARNING, "CAMLCodec::CAMLCodec libamcodec.so not found");
@@ -2332,9 +2363,6 @@ bool CAMLCodec::AddData(uint8_t *pData, size_t iSize, double dts, double pts)
     am_private->am_pkt.data_size = iSize;
   }
 
-  m_frameSizes.push_back(iSize);
-  m_frameSizeSum += iSize;
-
   am_private->am_pkt.newflag    = 1;
   am_private->am_pkt.isvalid    = 1;
   am_private->am_pkt.avduration = 0;
diff --git a/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.h b/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.h
index 05bd75278a..a78863b9dd 100644
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.h
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.h
@@ -129,9 +129,6 @@ private:
   PosixFilePtr     m_amlVideoFile;
   std::string      m_defaultVfmMap;
 
-  std::deque<uint32_t> m_frameSizes;
-  std::uint32_t m_frameSizeSum;
-
   static std::atomic_flag  m_pollSync;
   static int m_pollDevice;
   static double m_ttd;
```
media modules:
```diff
diff --git a/drivers/frame_provider/decoder/utils/vdec_input.c b/drivers/frame_provider/decoder/utils/vdec_input.c
index b32c760c..5631809a 100644
--- a/drivers/frame_provider/decoder/utils/vdec_input.c
+++ b/drivers/frame_provider/decoder/utils/vdec_input.c
@@ -307,6 +307,8 @@ void vdec_input_init(struct vdec_input_s *input, struct vdec_s *vdec)
 	input->block_id_seq = 0;
 	input->size = 0;
 	input->default_block_size = VFRAME_BLOCK_SIZE;
+	input->total_wr_count = 0;
+	input->total_rd_count = 0;
 	snprintf(input->vdec_input_name, sizeof(input->vdec_input_name),
 		 "vdec-input-%d", vdec->id);
 }
@@ -580,13 +582,13 @@ int vdec_input_get_status(struct vdec_input_s *input,
 	flags = vdec_input_lock(input);
 
 	if (list_empty(&input->vframe_block_list)) {
-		status->size = VFRAME_BLOCK_SIZE;
+		status->size = input->default_block_size;
 		status->data_len = 0;
-		status->free_len = VFRAME_BLOCK_SIZE;
+		status->free_len = input->default_block_size;
 		status->read_pointer = 0;
 	} else {
-		int r = VFRAME_BLOCK_MAX_LEVEL - vdec_input_level(input)
-			- VFRAME_BLOCK_HOLE;
+		int r = (input->default_block_size - VFRAME_BLOCK_HOLE) * 3
+			- vdec_input_level(input);
 		status->size = input->size;
 		status->data_len = vdec_input_level(input);
 		status->free_len = (r > 0) ? r : 0;
@@ -976,7 +978,11 @@ int vdec_input_add_chunk(struct vdec_input_s *input, const char *buf,
 	}
 	if (chunk->size > input->frame_max_size)
 		input->frame_max_size = chunk->size;
-	input->total_wr_count += count;
+	if (input->total_wr_count + (count + need_pading_size) < 0) {
+		input->total_wr_count -= input->total_rd_count;
+		input->total_rd_count = 0;
+	}
+	input->total_wr_count += count + need_pading_size;
 	vdec_input_unlock(input, flags);
 #if 0
 	if (add_count == 2)
@@ -1098,13 +1104,13 @@ void vdec_input_release_chunk(struct vdec_input_s *input,
 		input->last_comsumed_pts_u64 = chunk->pts64;
 	} else
 		input->last_comsumed_no_pts_cnt++;
-	block->rp += chunk->size;
+	block->rp += chunk->size + chunk->pading_size;
 	if (block->rp >= block->size)
 		block->rp -= block->size;
 	block->data_size -= chunk->size;
 	block->chunk_count--;
 	input->data_size -= chunk->size;
-	input->total_rd_count += chunk->size;
+	input->total_rd_count += chunk->size + chunk->pading_size;
 	if (block->is_out_buf) {
 		list_move_tail(&block->list,
 			&input->vframe_block_free_list);
diff --git a/drivers/frame_provider/decoder/utils/vdec_input.h b/drivers/frame_provider/decoder/utils/vdec_input.h
index edc07b53..5a516ff5 100644
--- a/drivers/frame_provider/decoder/utils/vdec_input.h
+++ b/drivers/frame_provider/decoder/utils/vdec_input.h
@@ -94,8 +94,8 @@ struct vdec_input_s {
 	ulong mem_handle;
 	void *swap_page;
 	dma_addr_t swap_page_phys;
-	u64 total_wr_count;
-	u64 total_rd_count;
+	int total_wr_count;
+	int total_rd_count;
 	u64 streaming_rp;
 	u32 swap_rp;
 	bool last_swap_slave;
```

kszaq:
@Portisch Thank you for the patch. I was finally able to check it today. Unfortunately, it made no difference...

Here's how to reproduce: play http://www.larmoire.info/jellyfish/media/jellyfish-140-mbps-4k-uhd-hevc-10bit.mkv and during playback, pause/resume a few times in a row until you clearly see frames jumping, lack of smoothness in jellyfish movement. Unfortunately it's not always reproducible but it should be possible to trigger the issue at some point...

Portisch:
@kszaq I finished now the rework of the AddData and GetPicture function. The functions are now updated to at least the 4.9 kernel handling. You might compare if something is changed for you now or maybe you can now filter your issue in more detail as the functions are much more simple than before.
https://github.com/CoreELEC/xbmc/commit/7734c1687efc4322056150b41c4d3ca23e3f7ec6

There is also a temporary change to render the hw frames (release them) before the GUI render. So the GUI renderer have now the time of one frame time to render. This solve most of render interfere, but not all.
End solution will be to decouple this two render into separate threads.
https://github.com/CoreELEC/xbmc/commit/ae821b0fffdc49940341922f261be111a05e7154

kszaq:
> @kszaq I finished now the rework of the AddData and GetPicture function. The functions are now updated to at least the 4.9 kernel handling. You might compare if something is changed for you now or maybe you can now filter your issue in more detail as the functions are much more simple than before.
> https://github.com/CoreELEC/xbmc/commit/7734c1687efc4322056150b41c4d3ca23e3f7ec6
> 
> There is also a temporary change to render the hw frames (release them) before the GUI render. So the GUI renderer have now the time of one frame time to render. This solve most of render interfere, but not all.
> End solution will be to decouple this two render into separate threads.
> https://github.com/CoreELEC/xbmc/commit/ae821b0fffdc49940341922f261be111a05e7154

This sound really great! I won't be able to give you feedback for at least 2 weeks but I hope @boot2k3 and your nightly testers will give it a good spin. 🙂

PR #10, opened 2023-07-05
Add Slovak translation

jose1711:
## Description
Add Slovak translation of Kodi including CoreELEC-related menus.

## Motivation and context
Currently CoreELEC menus are missing translation.

## How has this been tested?
Visual check and msgfmt.


Portisch:
Thx for your contribution but I think this is the wrong place for the language resourse.
I think it would be this place: https://github.com/xbmc/repo-resources/tree/omega

But TK will never accept a PR from a fork. Let us discuss internal how we can solve this.
Maybe we need to fork this [repo](https://github.com/xbmc/repo-resources)

Ricrdsson1:
Great idea, I've been thinking about it for a long time.  But I thought you wouldn't accept such a thing.  But if you included language packs with our modifications it would be great.  There are only a few texts, so it's not such a problem if it's not translated, but it would be better if it was.  Mainly, it would still be necessary to maintain the translation with the official translation so that it doesn't happen that someone gets pissed off about it and then the texts are missing...

jose1711:
> Mainly, it would still be necessary to maintain the translation with the official translation so that it doesn't happen that someone gets pissed off about it and then the texts are missing...

I suppose a workflow that would pull the latest translation from official Kodi translation and merge it with those few extra strings should be relatively straightforward to setup. In fact this was the way how I prepared this commit:
```
msgmerge kodi-core-kodi-main-sk_sk.po strings.po
```

Portisch:
We are quite unsure how we can manage this PR. The only solution will be we make a fork of the [repo](https://github.com/xbmc/repo-resources/tree/omega) and merge there this PR. But the problem is the language addons are directly embedded in kodi. So when you switch the language it directly download the addon and install it. But we have no man power right now for this if this is doable to use our repo instead the Kodi repo.

So I will close this PR for now and maybe come back to this item if the background is done to be able to merge such PR.

Thank you anyway for your support!

PR #11, opened 2023-11-24
Fall back to sw decode for MPEG-2 4:2:2

smp79:
When trying to play a MPEG-2 4:2:2 DVB stream it tried to use a hw decoder and it resulted in a black screen. This PR fixes it.


Portisch:
Is the hardware decoder really not able to handle it? Please provide a short video sample.

smp79:
[mpeg2_422.ts](https://www.dropbox.com/scl/fi/hxjv8hlmyfvgdvzlk4kmm/mpeg2_422.ts?rlkey=4k1fluyo81dh1px96wqr4fhkw&dl=1)

smp79:
No consumer-grade chip can hardware decode MPEG-2 4:2:2. Amlogic is no exception.
This issue didn't matter before because slower chips like GXL would choke on 1080p 4:2:2 anyway. But faster ones like A311D2 can play the above sample in sw with no problems.

Portisch:
Closed by https://github.com/CoreELEC/xbmc/commit/39600a430ce7aa7cd68235a7d4d4edc40404c2ad

PR #13, opened 2024-01-17
Update pt-br.po

xgrind:
## Description
translation correction


Portisch:
Please make the PR upstream as TK does maintain this file:
https://github.com/xbmc/xbmc/blob/master/addons/skin.estuary/language/resource.language.pt_br/strings.po

vpeter4:
Those strings comes from https://kodi.weblate.cloud/

PR #14, opened 2024-03-19
Change detection of EL Layer Type (MEL/FEL) for profile 7

cpm-code:
Change detection of EL Layer Type (MEL/FEL) for profile 7

## Description
Change from AMLogic detection of EL Layer Type (MEL/FEL) and restarting
the decoder to detecting in the BitstreamConverter and pass into decoder open.

## Motivation and context
Faster starting of playing content.
Do not skip over first few seconds of content.

## How has this been tested?
Tested against know library of content.

Better experience playing DoVi content.

## Types of change

- [ x] **Improvement** (non-breaking change which improves existing functionality)

## Checklist:

- [ ] My code follows the **[Code Guidelines](https://github.com/xbmc/xbmc/blob/master/docs/CODE_GUIDELINES.md)** of this project 

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
@@ -25,6 +25,7 @@
 #include "utils/StreamDetails.h"
 #include "utils/StringUtils.h"
 #include "utils/TimeUtils.h"
+#include "utils/BitstreamConverter.h"
 #include "ServiceBroker.h"
 
 #include "platform/linux/SysfsPath.h"
@@ -1816,8 +1817,6 @@ CAMLCodec::CAMLCodec(CProcessInfo &processInfo)
   , m_bufferIndex(-1)
   , m_state(0)
   , m_processInfo(processInfo)
-  , m_is_dv_p7_mel(false)
-  , m_dolby_vision_wait_delay(0)
 {
   am_private = new am_private_t;
   memset(am_private, 0, sizeof(am_private_t));
@@ -1854,7 +1853,7 @@ int CAMLCodec::GetAmlDuration() const
   return am_private ? (am_private->video_rate * PTS_FREQ) / UNIT_FREQ : 0;
 };
 
-bool CAMLCodec::OpenDecoder(CDVDStreamInfo &hints)
+bool CAMLCodec::OpenDecoder(CDVDStreamInfo &hints, DoviElType dovi_el_type)
 {
   m_speed = DVD_PLAYSPEED_NORMAL;
   m_drain = false;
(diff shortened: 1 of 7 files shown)


xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
```diff
@@ -2031,20 +2029,23 @@ bool CAMLCodec::OpenDecoder(CDVDStreamInfo &hints)
...
+        am_private->gcodec.dec_mode  = STREAM_TYPE_FRAME;
+        if (amdolby_vision_debug.Exists())
+          amdolby_vision_debug.Set("enable_fel 0");
+      } 
+      else // FEL and not determined
+      { 
+        am_private->gcodec.dec_mode  = STREAM_TYPE_STREAM;
+        if (amdolby_vision_debug.Exists())
+          amdolby_vision_debug.Set("enable_fel 1");
+      }
```
arthur-liberman: This if/else logic isn't needed. You can add `dovi_el_type` check in the first `if` that checks for the profiles.
If it's MEL, skip that if.

xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
```diff
@@ -2660,21 +2661,6 @@ CDVDVideoCodec::VCReturn CAMLCodec::GetPicture(VideoPicture *pVideoPicture)
...
-      m_dolby_vision_wait_delay--;
-      CSysfsPath is_mel{"/sys/module/amdolby_vision/parameters/is_mel"};
-      if (is_mel.Exists())
-      {
-        if (is_mel.Get<char>().value() == 'Y')
-        {
-          CLog::Log(LOGDEBUG, LOGVIDEO, "CAMLCodec::GetPicture: DoVi P{:d} MEL content detected, request to reopen decoder",
-            m_hints.dovi.dv_profile);
-          m_is_dv_p7_mel = true;
-          return CDVDVideoCodec::VC_REOPEN;
```
arthur-liberman, line 2674: How do we decide when to reopen the decoder with this removed and no new logic in its stead?

xbmc/utils/BitstreamConverter.cpp
```diff
@@ -279,19 +279,27 @@ static bool has_sei_recovery_point(const uint8_t *p, const uint8_t *end)
 #ifdef HAVE_LIBDOVI
 // The returned data must be freed with `dovi_data_free`
 // May be NULL if no conversion was done
-static const DoviData* convert_dovi_rpu_nal(uint8_t* buf, uint32_t nal_size, int mode)
+static const DoviData* process_dovi_rpu_nal(uint8_t* buf, uint32_t nal_size, int mode, int& dovi_el_type)
```
arthur-liberman: I don't think we should mess with Kodi's code too much here.
It may change in the future and we'll have to rebase it manually. I would think if there's a way to wrap this somehow in a dedicated method or separate block of code.

xbmc/utils/BitstreamConverter.cpp
```diff
@@ -995,27 +1009,26 @@ bool CBitstreamConverter::BitstreamConvert(uint8_t* pData, int iSize, uint8_t **
           m_sps_pps_context.idr_sps_pps_seen = 0;
       }
 
-      if (m_convert_dovi)
+      if ((m_convert_dovi || (m_check_dovi_el && m_dovi_el_type == 0)) && unit_type == HEVC_NAL_UNSPEC62)
```
arthur-liberman: Same here

xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
cpm-code: It is setting the enable_fel flag differently for FEL / MEL
If someone starts up and just plays an MEL would that flag be set?

Can leave out:
am_private->gcodec.dec_mode = STREAM_TYPE_FRAME;
but I left that for clarity for anyone reading the code, to determine the difference in handling in one place.

xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
cpm-code: The reopen here is just because of the detected MEL case - no longer need to reopen because of MEL is know up front.

PR #15, opened 2024-03-26
Enhanced VC-1 Acceleration Options

hceuterpe:
## Description
Provide enhanced user configurable options for VC-1 hardware acceleration, based on selection criteria

## Motivation and context
Currently there is no option to configure hardware acceleration for VC-1 playback. This presents issues in some instances such as hardware accelerated VC-1 decoding is subject to forced de-interlacing performed on even progressive content, resulting in considerable resolution loss for this content.  It appears hardware such as the S922X SoCs are powerful enough to perform software decoding of even 1080p24 content using the ffmpeg libraries for VC-1 playback.

Options are: Never, Always, HD and up, and Exclude 24p.  Defaults to Always (to retain existing behavior in lieu of user setting this option).

## How has this been tested?
Have tested multiple VC-1 content with all settings, including 24p, and interlaced content to ensure expected behavior.  As it appears no supported Amlogic hardware can perform sufficient and acceptable software decoding of interlaced VC-1, and therefore would still rely on hardware acceleration for these instances.

Provide more flexibility to the user so they can choose which options best suit their needs for VC-1 content playback, and provide the highest video quality and functionality.

## Types of change

- [X] **Improvement** (non-breaking change which improves existing functionality)
- [x] **New feature** (non-breaking change which adds functionality)

## Checklist:

- [X] My change requires a change to the documentation, either Doxygen or wiki

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -22961,6 +22961,11 @@ msgctxt "#39003"
 msgid "Accelerate h264"
 msgstr ""
 
+#: system/settings/settings.xml
+msgctxt "#39004"
+msgid "Accelerate VC1"
+msgstr ""
+
 #. Description of category "Library" with label #14202
 #: system/settings/settings.xml
 msgctxt "#39004"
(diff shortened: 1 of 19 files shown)


Portisch:
Please squash your commits to one or two (if need). But I think one commit is enough. Then we can start a review.
Please include spaces in code like ` != `, thx.

Please also make the PR to CE-21. When it works we can backport to CE-20 later.

PR #24, opened 2024-09-28
skin.estuary: Improved the player process info

frodo19:
## Types of change

## Checklist:

- [ ] My code follows the **[Code Guidelines](https://github.com/xbmc/xbmc/blob/master/docs/CODE_GUIDELINES.md)** of this project 

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamBluray.cpp
+++ b/xbmc/cores/VideoPlayer/DVDInputStreams/DVDInputStreamBluray.cpp
@@ -1077,7 +1077,7 @@ void CDVDInputStreamBluray::GetStreamInfo(int pid, std::string &language)
   if(!m_titleInfo || !m_clip)
     return;
 
-  if (pid == HDMV_PID_VIDEO)
+  if (pid == HDMV_PID_VIDEO || pid == HDMV_PID_VIDEO_EL)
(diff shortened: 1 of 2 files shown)


Portisch:
So this is how it looks now?
screenshot00005

In my opinion `Video codec` and `Video resolution` can be dropped as it's already be covered by `Video stream` and `Video decoder`.  `Video stream` is the input, `Display mode` the output format.

frodo19:
Do you should like someting this? 
gh

frodo19:
Hmm.. the pr gone again... its synced then automatically closed..

Portisch:
I did rebase xbmc, maybe this caused the close.
I am not sure about this `System FPS`. It might confuse people because its different to media FPS. Maybe rename it or drop it as it's not very useful anyway? Why should anybody care about the UI FPS during video playback?

frodo19:
> I did rebase xbmc, maybe this caused the close. I am not sure about this `System FPS`. It might confuse people because its different to media FPS. Maybe rename it or drop it as it's not very useful anyway? Why should anybody care about the UI FPS during video playback?

System fps is the current video output fps. That can be different to the source video fps. People like to see if that synced or not. 
But when you will , I can drop that too..

Portisch:
We think it's the UI fps and not related to video fps. It does also not match any time. So please drop it, yes.

Portisch:
We think it's the UI fps and not related to video fps. It does also not match any time. So please drop it, yes.

PR #38, opened 2026-04-13
fix 3D SBS/TAB/FP: always halve viewport in stereo modes

pannal:
When a display doesn't advertise 3D in its EDID, no `MODE3DSBS`/`MODE3DTB`/`MODE3DFP` resolutions get probed from `disp_cap_3d`. The whitelist picks a regular resolution without the 3D flag. The HDMI 3D signaling still fires (`3dlr`/`3dtb`/`3dfp` config write), so the display enters 3D mode, but the framebuffer isn't actually rendered as stereo because the viewport dimension halving is gated on the flag being present. Result: double image.

Fix: always apply the dimension changes (width halving for SBS, height halving for TAB, overscan for FP) when a stereo mode is active. The pixel ratio and flag adjustments stay conditional to avoid double-adjusting when a proper 3D resolution was selected.

Same bug is present on `aml-5.4-21.3` and `aml-5.15.196-22.0`.

### DIFF
--- a/xbmc/windowing/GraphicContext.cpp
+++ b/xbmc/windowing/GraphicContext.cpp
@@ -615,36 +615,39 @@ const RESOLUTION_INFO CGraphicContext::GetResInfo(RESOLUTION res) const
 
   if(m_stereoMode == RENDER_STEREO_MODE_SPLIT_HORIZONTAL)
   {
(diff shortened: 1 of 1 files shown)


Portisch:
Thanks, but project Amlogic-ng is closed.

PR #39, opened 2026-04-13
Fix interlaced A/V sync and video_rate on Amlogic

pannal:
## Summary

Three fixes for interlaced A/V sync on Amlogic.

The `video_rate` exception logic in AMLCodec.cpp had a few problems. The interlaced field-to-frame rate correction checked `hints.width == 1920` instead of `hints.interlaced`, so it missed SD interlaced content (576i50) and would incorrectly halve genuine 1080p50. It also used hardcoded NTSC pulldown math instead of deriving from the actual stream. Separately, the SD H.264 fps overrides (60fps/30fps to 23.976) fired for all containers including MKV/TS where stream-level timing is reliable — now limited to mp4/avi (`codec_tag != 0`) where container metadata actually lies.

In VideoPlayerVideo.cpp, the progressive fallback heuristic (which halves fps when picture duration consistently matches 2x the expected frame time) was false-triggering on MBAFF H.264 streams. MBAFF legitimately alternates between 20ms field output and 40ms full-frame output, and the 40ms frames looked like progressive to the heuristic, dropping fps from 50 to 25. Added a check against di0's `frame_format` sysfs — if the hardware deinterlace module says "interlace", trust it over the duration heuristic.

Also in VideoPlayerVideo.cpp: the di0 deinterlace pipeline buffers several fields before producing output, but Kodi captures PTS at V4L2 DQBUF (before the DI stage), so frames show up on screen ~240ms late. Audio ends up consistently ahead of video. This adds a 12-field-period timestamp shift when interlaced + AML HW decoder + di0 are all present.

### DIFF
--- a/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
+++ b/xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
@@ -1901,36 +1901,45 @@ bool CAMLCodec::OpenDecoder(CDVDStreamInfo &hints, enum ELType dovi_el_type)
   am_private->video_ratio      = ((int32_t)video_ratio.num << 16) | video_ratio.den;
   am_private->video_ratio64    = ((int64_t)video_ratio.num << 32) | video_ratio.den;
 
-  // handle video rate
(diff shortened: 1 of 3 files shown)


Portisch:
Thanks, but project Amlogic-ng is closed.

PR #42, opened 2026-04-17
VP: follow-up fixes for interlaced A/V sync on Amlogic

pannal:
Follow-up to #41. Several issues surfaced during wider testing, mostly around MBAFF H.264 content (1080i50 with progressive-coded macroblocks) and HEVC progressive content wrongly flagged as interlaced.

## DI pipeline latency now also covers ongoing frame release

The #41 approach shifted the start timestamp forward by 12 field periods, which fixed the initial A/V sync point but not ongoing frame release timing. At steady state video still lagged audio by ~240ms because the render manager's `displayLatency` was never told about the DI stage delay.

Added `SetDeinterlaceDelay`/`GetDeinterlaceDelay` on `CRenderManager`. The deinterlace delay is subtracted from `displayLatency` so frames are released earlier to feed the DI pipeline. The audio-side compensation is no longer done through `msg.timestamp` — that's now handled via direct audio sync corrections at DI state changes (see next section), which keeps the audio baseline independent of DI state. The initial interlaced offset falls out naturally from the `displayLatency` adjustment combined with SyncStream's normal alignment.

## Dynamic DI delay for mixed interlaced/progressive streams

MBAFF recap-style content genuinely alternates sections. The DI hardware adds ~240ms latency when deinterlacing but only ~60ms when passing progressive through. Added an audio sync correction channel (`IAEClockCallback::GetAndClearSyncCorrection`) that injects the exact pipeline-latency delta into ActiveAE on interlace↔progressive transitions, bypassing the averaging window that would otherwise mask sub-threshold (<100ms) corrections.

Field counts are exposed as advancedsettings (`dipipelinefields=12`, `diprogressivefields=3`) for tuning without rebuilding.

## MBAFF edge cases for the progressive fallback heuristic

The fallback added in #41 needed to be more careful with MBAFF, since MBAFF legitimately mixes 20ms field output and 40ms frame output. Four changes:

- Require positive `"progressive"` from vfmt rather than `"not interlace"`. The old check passed when vfmt was still empty/null (DI not ready yet), and with only 4 consecutive frames needed the fallback could fire at playback start and halve fps.
- Don't let transient DI `"progressive"` readings (reflecting the current macroblock coding type, not the stream) clear the demuxer's interlace flag.
- When the demuxer has flagged the stream as interlaced (`CODEC_INTERLACED`), skip the fallback entirely and trust the demuxer over the duration heuristic.
- When the demuxer says interlaced, also skip `CalcFrameRate`'s halving of `m_fFrameRate` (50 → 25). MBAFF progressive sections make the PTS tracker measure a 25fps cadence and the halving caused visibly "laggy" motion in subsequent progressive scenes.

## 29.97/25fps progressive wrongly treated as interlaced

`OpenVideoStream` was unconditionally doubling 29.97→59.94 (and 25→50) and marking interlaced regardless of `hints.interlaced`. This broke HEVC progressive content (e.g. 1440x1080 29.97p) because di0 is not in the HEVC pipeline, so the vfmt-based correction never fires and the content stayed stuck as interlaced 58.94.

Fix: gate the interlace assumption on `hint.interlaced`. Also fixes a double-division bug in the progressive retry fallback (59.94 → 29.97 → 14.985).


Portisch:
I am not sure if this PR cause now a be early audio than video on 1080i50hz. But I see now it's not in sync anymore. It was in sync before.

pannal:
Hmm which device did you test this on? It might be that the DI pipeline has a different delay than on the S922XJ. Try adjusting the advancedsetting accordingly.

PR #46, opened 2026-05-13
DVDVideoCodecAmlogic: reset DoVi start decode after seek

Tristan1019-user:
## Summary

Reset the Amlogic bitstream converter's start-decode state after reset/seek discontinuities for Dolby Vision streams.

## Why

CDVDVideoCodecAmlogic::Reset() already resets start-decode state for H.264 and VVC cases that need a fresh keyframe/start point after a discontinuity. Dolby Vision playback also uses CBitstreamConverter, and stale start-decode state can leave the decoder resuming from an invalid point after seeks.

This extends the existing reset path to Dolby Vision, FEL, and MEL streams while preserving the converter metadata and cached bitstream state.

## Scope

This is the source-side replacement for the incorrectly-targeted CoreELEC/CoreELEC PR. It intentionally does not include packaging changes, AVDV UI work, or Fire TV Cube device-specific patches.

## Validation

- git diff --check HEAD~1..HEAD passed.
- Downstream CoreELEC 22 build using this behavior was used during Dolby Vision/FEL playback and seek testing on Fire TV Cube 2nd Gen.


Portisch:
This can not work as start flag is never set on h265.

PR #48, opened 2026-05-13
Use three decimal places for display mode

CE-Repo:
Before:
591530411-24f49e41-da06-4ecc-99a1-46c487962fab

After:
591530442-dcd6b4ae-0c98-4c8c-b4ec-1bb22a36699e


Portisch:
Your PR will still show 59.940...

PR #56, opened 2026-05-28
windowing/amlogic: fix SD output aspect ratio

turley:
## Description

`aml_mode_to_resolution()` was setting `fPixelRatio = 1.0f` for every mode, including 720x480 and 720x576 - which are anamorphic and need a non-square pixel ratio. This adds a special case for those two raster sizes so `fPixelRatio` reflects the actual display aspect (32:27 and 64:45 respectively) rather than the raster aspect. All other modes are unchanged.

## Motivation and context

SD content played at native 720x480 / 720x576 HDMI output is visibly stretched horizontally. Anamorphic 16:9 DVDs are the worst case (~40% too wide - circles become ovals, people look squashed), but 4:3 SD content is wrong too (~19% stretch on a 16:9 display).

I noticed this while playing DVD rips with SD modes whitelisted (I do this to let a downstream video processor upscale content from its original size).

The root cause is in `AMLDisplay.cpp`: `aml_mode_to_resolution()` sets `fPixelRatio=1.0` for every mode, but the 720x480 / 720x576 framebuffers used for SD HDMI output are anamorphic. The kernel HDMI driver signals those modes as 16:9 (VIC 3 / VIC 18 by default), so the sink stretches the 720-wide raster to fill 16:9 on the wire. With `fPixelRatio=1.0`, Kodi's renderer also treats the framebuffer as 3:2 square pixels - so for a 16:9 source it adds its own 16:9 letterbox inside the surface (`/sys/class/video/axis = "0 37 719 441"` - a 720x405 box centered in a 720x480 frame). The two stretches stack and you get the ~40% distortion. For a 4:3 source the renderer pillarboxes to 4:3 inside that same 3:2-treated surface, and HDMI then stretches the whole frame to 16:9, producing the ~19% case.

The fix assumes the HDMI sink displays the 720x480 / 720x576 surface as 16:9, which is the Amlogic default; a user who manually forces 4:3 HDMI signaling via `/sys/class/amhdmitx/amhdmitx0/aspect` (unusual) would need different math.

The same underlying bug appears to exist in upstream Kodi's GBM windowing path (`xbmc/windowing/gbm/drm/DRMUtils.cpp`). I plan to submit a separate PR there. Related upstream issues: xbmc/xbmc#22351, xbmc/xbmc#23434.

## How has this been tested?

Verified on Ugoos AM6B+ (S922X, coreelec-22 nightly with this change cherry-picked into the kodi package), both 16:9 and 4:3 cases.

I've prepared four test fixtures, each containing a circle inscribed in a square, MPEG2 interlaced TFF - matching what MakeMKV produces from a DVD:

- 480i 16:9 (NTSC anamorphic): https://shared.luminationlabs.com/anamorphic-480i-16x9-ybU0XHLtcMavsX6qkYZ4ynU6V
- 576i 16:9 (PAL anamorphic): https://shared.luminationlabs.com/anamorphic-576i-16x9-cLG4jbI6r2DVtXxy4GcbCilcv
- 480i 4:3 (NTSC fullscreen): https://shared.luminationlabs.com/fullscreen-480i-4x3-fjfYHm0CNrtJ0PzAPzygVBJyw
- 576i 4:3 (PAL fullscreen): https://shared.luminationlabs.com/fullscreen-576i-4x3-w90i6bwQjoFaHwWnes1ggoMjq

The source is interlaced because that's what real DVD rips are; Kodi will deinterlace it to whichever SD mode is in the whitelist (typically the progressive variant). The framebuffer on the wire is anamorphic either way (VIC 3 / VIC 18 for progressive, VIC 7 / VIC 22 for interlaced - all with the same SAR), so the bug shows up regardless. On a 16:9 display the circle should be round and the square should be square; pre-fix they're visibly distorted.

In Settings > System > Display, make sure your whitelist has at least one 480-line entry (e.g., `720x480p 59.94Hz`, `720x480p 60Hz`) and/or 576-line entry. Then play a fixture. While playing, on the box:

    cat /sys/class/video/axis

For 16:9 sources:

- Pre-fix: `0 37 719 441` (or `0 47 719 528` for 576-line)
- Post-fix: `0 0 719 479` (or `0 0 719 575`) - full frame

For 4:3 sources, both pre- and post-fix fill the surface vertically, but the renderer's pillarbox width is different (correct after the fix).

SD content played at native 720x480 / 720x576 HDMI output displays with the correct aspect ratio. Anamorphic 16:9 DVDs no longer appear ~40% too wide, and 4:3 SD content no longer appears ~19% too wide on 16:9 displays.

Users who've previously run Video Calibration on 480p / 576p will have a saved `<pixelratio>` entry in `guisettings.xml` that overrides the new default. They'll need to reset calibration for those modes (Settings > System > Display > Video calibration > Reset) to pick up the fix. Users who never calibrated SD modes are unaffected.

## Screenshots:

| Source                  | Before / after                                                                                                |
| ----------------------- | ------------------------------------------------------------------------------------------------------------- |
| 720x480 anamorphic 16:9 | compare-anamorphic720x480 |
| 720x480 4:3             | compare-fullscreen720x480 |
| 720x576 anamorphic 16:9 | compare-anamorphic720x576 |
| 720x576 4:3             | compare-fullscreen720x576 |

Note: These are photos (not screenshots), so there's slight perspective from the camera angle, and the anamorphic shots are scaled to screen height by an external video processor. Aspect ratio is preserved in all cases.

## Types of change

## Checklist:


Portisch:
I can see it fix this media for 16:9 display in 576p50hz display mode.
But what is about 4:3 displays? The result will be a squezzed?

Also I think the fix will be more correct here:
https://github.com/CoreELEC/xbmc/blob/aml-5.15.196-22.0/xbmc/cores/VideoPlayer/VideoRenderers/BaseRenderer.cpp#L393

Like for "normal":
```c
    CDisplaySettings::GetInstance().SetZoomAmount(1.0);
    CDisplaySettings::GetInstance().SetPixelRatio(1.0f / ((16.0f / 9.0f) / sourceFrameRatio));
```

But this match again only if display is 16:9 but resolution and video is 4:3.
So I think kodi need to know first the display aspect ratio.

turley:
Thanks for picking this up and addressing the 4:3-display case. The EDID check is a nice touch.

PR #63, opened 2026-07-08
Windowing: Fix GUI tearing by moving CreateGPUFence after eglSwapBuffers

simonhust:
Fix GUI tearing.


Portisch:
I will try to add back the GPU fence in AML windowing as it was dropped lately.

But this PR is invalid as it touch `‎xbmc/windowing/gbm` what is not used by AML.
For GBM windowing please go upstream: https://github.com/xbmc/xbmc

PR #65, opened 2026-07-18
AML: re-decide the HDMI colour on Dolby Vision enable/disable

allolive:
## Description

Amlogic only. Re-decides the HDMI colour attribute whenever Dolby Vision turns on or off, so DV runs on its tunnel base instead of the stale SDR one. This is a colour re-decide at the same timing, not a refresh or mode switch, so it does not touch the refresh-rate switching logic.

## Motivation and context

Two cases, same root cause:

- **4K50/60:** DV came up RGB 12-bit, which is over HDMI 2.0 bandwidth, so the sink showed no signal.
- **Fractional refresh (24.000 to 23.976):** a fractional change keeps the same VIC and only flips FRAC_RATE_POLICY, so going from a 24.000 title to a 23.976 one (or back) does not re-decide the DV base, and the second title comes up at the wrong depth (10-bit instead of the tunnel base).

Both happen because decide_color_attr runs before the DV EOTF is set, and a DV start or a fractional flip at an unchanged timing has no modeset to re-run it. Watching the DV enable flag and re-committing covers all of them at any resolution.

## How has this been tested?

Ugoos AM6B Plus (S922X, HDMI 2.0), CoreELEC 22. DV P7 titles in both TV-led and player-led:

- forced 4K60: TV-led 4:4:4 8-bit, player-led 4:2:2 12-bit.
- 24.000 to 23.976 and back, title to title: base held across the fractional switch in both directions and both modes.

Verified on the wire at /sys/class/amhdmitx/amhdmitx0/config, no link drop.

Dolby Vision displays correctly at 4K50/60 and after a fractional refresh change, where before it was no signal or the wrong colour depth.

## Types of change

## Checklist:

> **Depends on** CoreELEC/common_drivers#31 — the driver picks the DV base at decide time; this side only makes decide run at the right moment.


Portisch:
Closed as related to https://github.com/CoreELEC/common_drivers/pull/29#issuecomment-5031748837

PR #67, opened 2026-08-07
Playerprocessinfo: add live Dolby Vision and HDR metadata infolabels

matthane:
# Live Dolby Vision and HDR metadata InfoLabels

Adds 58 `Player.Process(...)` InfoLabels exposing the Dolby Vision RPU metadata, the HDR10 static metadata and the HDR type of the playing stream, updated per frame where the metadata is per frame. No new dependencies; `libdovi` is already a Kodi dependency on CoreELEC 22.

`processDoviRpu` already runs on every unspec62 NAL, but early-exits once the enhancement layer type has been probed unless the profile 7 to 8 conversion or the zero level 5 rewrite is on. So today only the first RPU of a stream is parsed with both off, and with a rewrite on every RPU is parsed and the result thrown away. This drops the early exit so extraction always runs, reads the metadata back after whatever rewrites applied, and publishes it to `CDataCacheCore` where the existing label plumbing reaches it. The same pass picks up the HDR10 mastering display and content light level SEI, which was not being read at all. Always parsing is the real cost; it is bounded by there being at most one RPU per frame and by a byte compare cache on the input NAL, measured below.

## Labels

```
video.dovi.         has.config has.header version.major version.minor profile level
                    rpu.present el.present bl.present bl.signal.compatibility el.type
                    dual.track codec.fourcc codec.string meta.version
                    source.{min,max}.pq source.{min,max}.nits
                    has.l6 l6.max.cll l6.max.fall l6.min.lum l6.max.lum
                    l1.{min,max,avg}.{pq,nits} has.l5 l5.{left,right,top,bottom}.offset
video.source.dovi.  profile el.present el.type bl.signal.compatibility
video.              hdr.type[.raw] source.hdr.type[.raw] source.additional.hdr.type[.raw]
video.hdr.          has.mdcv min.lum max.lum colour.primaries has.cll max.cll max.fall
video.              bit.depth color.space color.range color.primaries
                    color.transfer.characteristic render.pts
```

The `l1.*` and `l5.*` values describe the frame on screen. `has.l5` reads 1 exactly when the RPU carried level 5 metadata. The rest of `video.dovi.*` is published on the first parsed RPU and reads defaults until `video.dovi.has.header` turns 1; the two codec labels are the exception and go live at decoder open.

## Design notes

Labels report the stream, not the display chain: `video.dovi.*` describes the bitstream handed to the decoder, which is what Kodi can observe. Nothing reads or writes sysfs.

Source and output labels come in pairs and the disagreement is the useful part. With the profile 7 to 8 conversion on, `video.source.dovi.profile` reads 7 while `video.dovi.profile` reads 8. With "Tone map HDR to Dolby Vision" on, the VS10 engine rewrites `hints.hdrType` before the codec sees the hints, so `video.hdr.type` reads dolbyvision for an HDR10 file; `CDVDStreamInfo` gains a `sourceHdrType` that only the demuxer assignment writes, so the pair disagreeing is how a skin knows the box synthesised it. Same shape when a display without HDR10+ makes CoreELEC strip it: source reads hdr10plus, output hdr10.

Per frame metadata is keyed by packet timestamp and resolved against the render timestamp `CRenderManager::Render()` publishes. Exact key matching never hits, because the Amlogic decoder returns a reconstructed clock on a rounded 90 kHz grid offset from the container's and separating by 64 microseconds per second of playback (measured on S905X5-J, detail on request), so a lookup falling back to the newest entry returned the most recently decoded frame and the labels led the picture by the decode queue depth. `findNearest` resolves to the closest key instead, which stayed frame accurate through a full length feature.

The cache compares NAL bytes, not metadata values. On profile 7 FEL the RPU NAL differs on every frame because it also carries the enhancement layer composer coefficients, which vary frame to frame even inside a static shot, while the display management values inside it, L1 in particular, only change on a cut. FEL is therefore the cache's worst case and is where these came from: over a 317 second steady state window on a 4K profile 7 FEL title, 24.0 parses per second, which is the 23.976 frame rate exactly and is itself the evidence that no two consecutive FEL RPUs matched byte for byte, plus 711 cache hits per second absorbing the decoder's packet re-feeds under backpressure. Whole process CPU sat at 30 to 45% of one core with debug logging on, and the parses are 24 microsecond-scale libdovi calls per second, well under 0.1% of a core. Single layer profiles carry no per frame composer payload and should cache across frames too, but that was not measured and the worst case is the right basis for the cost argument.

`processDoviRpu` now returns data owned by the converter, valid until the next call or `Close()`, and the three call sites copy immediately. That is what lets the cache hold the rewritten RPU alive.

## Two standalone bug fixes

Both are pre-existing defects, included because the new code depends on them, and both stand alone.

`HevcSei: bound the sei payload against the available bytes` compared a byte count against `br.AvailableBits()`, so a truncated NAL could pass carrying eight times more payload than it holds and let a reader run past the buffer. The fix is `/ 8` plus its comment, and it cherry-picks onto `dbcb4129` unmodified.

`BitstreamConverter: detect HDR10+ while it is being removed` drops a `!m_removeHdr10Plus` term that disabled HDR10+ detection on exactly the displays that strip it, so an HDR10+ file was reported as plain hdr10 everywhere including the OSD. The emitted bitstream is byte identical either way. Taken ahead of this series it needs a one-line context resolution, since the HDR static metadata commit inserts a call directly above the edited condition.

## Known limitations

- Profile 7 dual track updates L1 about 50 to 100 ms late after a cut. The keying was checked and is consistent with what the decoder is handed for the combined access unit; the residual comes from the kernel's own base layer pts tracking on dual track FEL, which the CE21 lineage addressed with a kernel patch this series deliberately does not carry.
- HLG signalled only through the alternative transfer characteristics SEI reads as SDR in `DetermineHdrType`, so the HDR type labels are empty. Pre-existing; changing it moves VS10 routing.
- Software decode does not publish these. A producer was written and reverted: scan-time `CProcessInfo` instances would clobber the playing stream through the cache singleton.
- `video.source.additional.hdr.type` is only written when HDR10+ is detected, so switching away from an HDR10+ stream mid-playback leaves it stale until file close.
- `CBitstreamConverter::Close()` resets only part of the Dolby Vision state; harmless while a fresh converter is built per open.
- `CDVDStreamInfo::sourceHdrType` is not in `Equal()`; safe today since it is written with `hdrType` from the same demuxer stream.
- `video.dovi.has.header` is structurally always 1 once an RPU is parsed. Kept as the "metadata is live" gate the other labels use.
- `video.dovi.codec.fourcc` reads `----` when the container has no codec tag, in practice MKV, so the codec string reads like `----.07.06`. A profile-derived guess would misrepresent a stream attribute.

## Testing

Ugoos AM9 Pro (S905X5-J), CE22 built from this branch, labels read live through an overlay addon and over JSON-RPC. Covered profile 5, profile 8.1, profile 7 MEL, profile 7 FEL single and dual track, HDR10+ without Dolby Vision, plain HDR10 and SDR. Verified by eye that L1 tracks scene cuts, freezes on pause and re-syncs after a seek, that the zero L5 setting is reflected honestly, and that per frame values match the frame on screen rather than running ahead of it. Worst case parse cost measured as above. Every commit builds standalone on the Amlogic target. A scripted four-list audit checks name, enum, resolver case and documentation block agree both ways; its baseline records four pre-existing documentation defects in the label table that deserve their own small PR.

## Enum headroom and an HDR10+ follow-up

These 58 labels sit at `PLAYER_PROCESS_START + 34..91`, leaving 8 slots before `ADDON_INFOS_START = 1600`. HDR10+ per frame labels are designed and ready as a follow-up but need 6 to 9 of those, so they are blocked on moving that boundary. Moving it looks safe: three derived constants consumed in four places, no range checks, IDs never serialised, skins and JSON-RPC address by string, and the Kodi 21 lineage did the same move; for Kodi 22 the target would be 1800, since RETROPLAYER now occupies 1700 to 1714. That is a core layout change, so it is flagged here rather than smuggled into an Amlogic scoped PR.

## Lineage

The framework is cpm's work from late 2024, which reached the CoreELEC 21 forks through an avdvplus sync and then pannal's p3i build. This ports the mechanism to Kodi 22 on CoreELEC's tree, drops everything with a kernel dependency, rewrites the expression to match this tree, and fixes the metadata lookup and the conditional pts stamping, dropping the L5 detector field with the detector.


quietvoid:
It should be preferable to figure out a way to have a callback on every FFmpeg parsed frame instead of messing around with the bitstream.
All of the info is already reachable in frame side data.

matthane:
> It should be preferable to figure out a way to have a callback on every FFmpeg parsed frame instead of messing around with the bitstream.
> 
> All of the info is already reachable in frame side data.

Thanks for your feedback. I will investigate that approach.

matthane:
From what I can see, the Amlogic path doesn't have an FFmpeg frame to hang the callback on: the demuxed bitstream goes straight to the hardware decoder through amcodec, so no libavcodec decoder runs. AV_FRAME_DATA_DOVI_METADATA looks to be attached only by the libavcodec decoders, the dovi_rpu BSF parses but doesn't export metadata, and the only packet level DV side data is AV_PKT_DATA_DOVI_CONF.

Everywhere FFmpeg does decode, the approach works as is, and mainline already reads the parsed AVDOVIMetadata off the frame for the EL type (xbmc#27914). Extending that consumer to the full struct and the InfoLabels upstream looks like the right home for the generic part of this, with no second parse anywhere, and it lines up with what @Portisch advised on the forum. That would leave only a small Amlogic producer in the CE tree.

For that producer, the current thinking is to export the parsed metadata as packet side data in the CE FFmpeg tree, the same pattern as the existing AV_PKT_DATA_DYNAMIC_HDR10_PLUS_RAW patch. That keeps the parsing inside FFmpeg and Kodi only ever reads side data. Happy to hear if you see a problem with that shape.

EDIT: Dug into the side data shape further and it has two problems on this pipeline. Packet side data is only populated by the internal FFmpeg demuxer, so streams coming through binary addon inputstreams (inputstream.adaptive and friends) would never carry it. And a demux time parse reports the source stream, not what is actually sent to the decoder after the bitstream converter has rewritten the RPU (profile conversion, L5 zeroing on the setups that do that).

So the better shape looks like a small parse-only public API in the CE FFmpeg tree, two additive files wrapping the RPU parser the decoders already use, called from the Amlogic codec on the post-conversion elementary stream. Parsing stays inside FFmpeg with no duplicate implementation, Kodi still only ever reads the parsed struct, and every input path is covered.

matthane:
Closing this in favor of the upstream approach discussed above. The generic part (frame metadata cache and info labels, fed from FFmpeg frame side data) is now proposed directly to Kodi as https://github.com/xbmc/xbmc/pull/28896. If the upstream PR lands, I'll work on the CoreELEC small Amlogic producer feeding the same path, which will come as a much smaller follow-up.

PR #68, opened 2026-08-10
Add a CoreELEC label registry and raw DV/HDR side data for add-ons

matthane:
## Description
This adds a small label registry so that CoreELEC info labels can live in CE owned files instead of rows in the Kodi label tables, and moves the existing CE labels onto it. On top of that, the Amlogic codec publishes the stream's raw Dolby Vision and HDR metadata under one new label, Player.Process(video.sidedata): the DV RPU, HDR10+ SEI, static HDR SEIs and DV configuration record, latched from the demuxer packets before any conversion can strip them, base64 encoded in one small JSON object, and sequenced per pts against the render clock so a poll tracks the frame on screen. The latch is codec agnostic and takes both the payloads carried in band and the ones the container hands over as side data. Core parses none of it. The byte contract is documented in the label's dox entry.

## Motivation and context
The goal is to lower the maintenance cost of CE labels, not add to it. The shared Kodi files carry two small hooks, everything else lives in CE owned files, labels resolve by name instead of competing for id slots, and a rebase check script guards the CE ranges. Any future side data type is just a new key in the JSON object.

Parsing lives outside the tree in a module add-on that binds libdovi and the platform's libavutil directly: https://github.com/matthane/script.module.sidedata. Ideally, this script module could live in a CE repo for addons to easily use as a dependency.

System.BatteryLevel moves back to its upstream id with linuxver served from a CE id instead, and the System Info presentation from the old patch is re-applied in its own commit so it can be kept or dropped independently.

## How has this been tested?
On a Ugoos AM9 Pro (S905X5-J) running a CoreELEC 22 test build. Playback of all current supported DV profiles, including profile 10 AV1, with the parsed output cross-checked byte for byte against dovi_tool and FFmpeg on real titles, shown by a diagnostic overlay: https://github.com/matthane/script.dvhdr.labels.diagnostic/releases/tag/v3.0.3 HDR10+ arriving as container side data was checked the same way.

Add-ons can read the stream's live Dolby Vision and HDR metadata during playback. There should be no visible change otherwise.

## Types of change
- [X] **Improvement** (non-breaking change which improves existing functionality)
- [X] **New feature** (non-breaking change which adds functionality)

## Checklist:
- [X] My change requires a change to the documentation, either Doxygen or wiki
- [X] I have updated the documentation accordingly


Portisch:
For the battery item:
Please complete revert https://github.com/CoreELEC/xbmc/commit/84b93f406c2abc03923c809cc1e525e81b8a96bf and then readd the Linux version like you reworked in this PR.

I will than drop the old commit completely afterwards.

matthane:
Done. The old patch is now fully reverted in its own commit and the Linux version comes back separately on a CE id. The revert also restores the kernel suffix on the OS version line and drops the AMLogic title text since those came from the same patch.

matthane:
While this is open, I've been experimenting on top of this branch with even more DV labels (L2/L8 trim passes for example) and HDR10+ dynamic metadata labels. With those added AMLFrameMetadata.h grows past 2000 lines, so I think it makes sense to create more of a formal framework per format. AMLFrameMetadata.h keeping the registry and shared plumbing, with the DV and HDR10+ parsing/rendering moving to their own headers (AMLDoviMetadata.h, AMLHdr10PlusMetadata.h). Then, we would have a cleaner structure that I think would be easier for any future additions too.

Should I add that work and update this PR with it? Or is it easier to keep this one as is, and send a follow up with those changes pending this one?

Portisch:
I am still offline so can't check much.

I would prefer to transfer the raw HDR meta data to a add-on as byte array. So all parsing and structs can be moved outside. This is much easier to maintain as when in main core code.

CE kernel just do the same. So pass the raw data to a buffer of the addon would be best.

matthane:
Ah ok, yeah I see what you mean. So then it would be passing the raw bytes (base64) in something like player.process(video.dovi.rpu) and player.process(video.hdr10plus.sei). And then an addon could parse them and open up the specific metadata as fields and be used by other addons as a dependency. That would definitely clean things in the core base!

I'll rework this PR to strip the parsing and structs down to just the raw pass-through.

Portisch:
Yes, or just raw data with the info what side data it includes. So the raw data can be typeless transfered to the add-on.
I am not sure but TinyPPi use same approach to transfer data to the skin?

matthane:
Ok, I'll experiment and see what works with minimal core changes.

TinyPPI currently uses a binary (hdrprobe) to oneshot inspect each file on playback and get the hdr/dv metadata from it's output json. But, it's all static metadata because of this.

Portisch:
I think we can live with it if the bitstream converter provides the current frame data through an interface, as this would require minimal code changes. However, I’m not sure how this will work when the FFmpeg BSF parser is used.

Would this be done by extracting the side data?

matthane:
Basically, anything FFmpeg already delivers as side data (the HDR10+ block additions from MKV, the DV config record) is just passed through. For the in-band payloads (RPU NAL, SEIs) the codec grabs the bytes from the raw demux packet before any conversion runs, so it doesn't matter whether the bitstream converter or an FFmpeg BSF processes the packet afterwards. Nothing needs an interface on the converter at all. It all comes out as one label with the payloads tagged and base64'd, and the add-on does the rest.

I think it will be easier to discuss or review further once I've reshaped the PR to simplify it accordingly. Thanks for your feedback so far, I think it will be much more elegant this way.

matthane:
Ok, I've updated the title, description, and commits now. The biggest chunk is b85b834a6 because it holds how we gather the raw metadata from the packets, before the bitstream converter can strip or rewrite it, and package it for the new label.

Edit: While adapting to the new base I found the HDR10+ kernel feed broke with the ffmpeg 9.0 bump. The old raw side data patch was dropped in favor of AV_PKT_DATA_DYNAMIC_HDR_SMPTE_2094_APP5, but that type carries SMPTE 2094-50, which is a different format. Real HDR10+ mkv block additions now arrive parsed as AVDynamicHDRPlus under AV_PKT_DATA_DYNAMIC_HDR10_PLUS, so the checkin never fires. The new first commit 768c92a  restores the feed by querying the right type and rebuilding the raw T.35 with ffmpeg's own serializer, the same way the ffmpeg mkv muxer writes it.

Portisch (commented):
Is it ok for you when I squash this fix to https://github.com/CoreELEC/xbmc/commit/c4b6ceb2c0dcc005e9fa39402e1f6911b0d7ec51 ?
I tested it and HDR10p pop up again on my VP9 sample.

And VP9 with HDR10p is also handled correct?
I see only h265 and AV1 handling. Maybe make it more "codec independent"?

matthane:
Sure, go ahead and squash it, that's the cleanest place for it. And good catch on VP9! I always forget about it. I've removed that gate so it's fully codec independent, and verified with a VP9 HDR10+ sample that the label now carries its per-frame payloads too. Will push together with the rebase once you've squashed.

xbmc/guilib/guiinfo/CEGUIInfoProviders.h
```diff
@@ -147,9 +181,11 @@ inline void Register(CGUIInfoManager& infoManager)
   registry.Add("player.process(amlogic.eoft_gamut)", CE_PLAYER_PROCESS_AML_EOFT_GAMUT);
   registry.Add("player.process(audiochannelssink)", CE_PLAYER_PROCESS_AUDIOCHANNELS_SINK);
   registry.Add("system.linuxver", CE_SYSTEM_LINUX_VER);
+  registry.Add("player.process(video.sidedata)", CE_PLAYER_PROCESS_VIDEO_SIDEDATA);
```
Portisch, line 182: `CE_PLAYER_PROCESS_VIDEO_SIDEDATA` is not defined?

Portisch (commented):
Instead "move" please revert complete [Add Amlogic display mode, EOFT and colourimetry](https://github.com/CoreELEC/xbmc/commit/9c86cd00798af12f779e5f463258f92f62110e45) and [alsa: provide ALSA output channel layout](https://github.com/CoreELEC/xbmc/commit/793587be5e09ca832a09218b6635d8dfa2893136) and make a new implemention.

Otherwise it will continue fail on rebase as it's still be applied to core code and then removed again.

Once merged I will drop whole commit and revert on a rebase afterwards.

xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLFrameMetadata.h
```diff
@@ -0,0 +1,559 @@
...
+  static constexpr double PTS_TOLERANCE = 1000.0; // DVD_TIME_BASE is microseconds
+  static constexpr double BACKWARD_JUMP = 5000000.0;
+  // deep enough that after compaction only per frame churn can overflow it
+  static constexpr size_t MAX_DEPTH = 512;
+  static constexpr size_t REORDER_MARGIN = 64;
+
+  std::map<double, AMLFrameMetadata> m_queue;
+};
+
+constexpr uint32_t CE_PLAYER_PROCESS_VIDEO_SIDEDATA = PLAYER_PROCESS_START + 34;
```
Portisch: `CE_PLAYER_PROCESS_VIDEO_SIDEDATA` is definded here?

It should rest at the same place as the others.

xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLFrameMetadata.h
```diff
@@ -0,0 +1,559 @@
...
+        payload += 255;
+        ++p;
+      }
+      if (p >= end)
+        break;
+      payload += rbsp[p++];
+      if (payload > end - p)
+        break;
+
+      if (type == 4) // user_data_registered_itu_t_t35
```
Portisch: Please use enum, preferred if already be defined in source.
Easier to read in future. Otherwise each type must be searched again when checking the source.

Same with the NAL units, 39, 62,...

xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAmlogic.cpp
```diff
@@ -448,6 +507,30 @@ bool CDVDVideoCodecAmlogic::AddData(const DemuxPacket &packet)
...
+    if (m_dualLayer && m_streamMeta.structure != "dt-dl")
+    {
+      m_streamMeta.structure = packet.isDualStream ? "dt-dl" : "st-dl";
+      m_pendingMeta.structure = m_streamMeta.structure;
+    }
+
+    // latch from the original demuxer payload, before Convert() can strip or
+    // rewrite it. Dual-track streams are latched from the base layer at pair
+    // completion: the EL carries its own static SEIs with different values
+    if (m_hints.codec == AV_CODEC_ID_HEVC && !packet.isDualStream)
```
Portisch: Better use `m_hints.hdrType` to check what side data is handled.
This can avoid missing codec id in the switch.

xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAmlogic.cpp
```diff
@@ -577,6 +700,54 @@ bool CDVDVideoCodecAmlogic::AddData(const DemuxPacket &packet)
...
+
+  AMLFrameMetadata meta;
+  if (m_metadataSequencer.Consume(target, meta))
+  {
+    CAMLFrameMetadataStore::GetInstance().Publish(m_metadataToken, meta);
+    if (!m_metaLeadLogged)
+    {
+      m_metaLeadLogged = true;
+      CLog::Log(LOGDEBUG, "{}: frame metadata pts lead {:.0f} ms", __MODULE_NAME__,
+                (m_lastCommitPts - target) / 1000.0);
```
Portisch: Please follow code style like: `(renderPts - nextFramePts) / DVD_TIME_BASE`

xbmc/guilib/guiinfo/CEGUIInfoProviders.h
matthane: Done, moved to CEGUIInfoRegistry.h next to the others.

xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLFrameMetadata.h
matthane: Done, moved to CEGUIInfoRegistry.h next to the others.

xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLFrameMetadata.h
matthane: Done. The HEVC NAL and SEI payload types are enums now. As far as I can tell ffmpeg doesn't install sei.h or hevc.h, so those are defined locally, and the AV1 OBU types come from obu_util.h the same way AMLCodec already uses it.

xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAmlogic.cpp
matthane: Reworked to check hdrType, with a codec switch picking the parser and a default arm so a new codec can't slip through unhandled. I went with hdrType != NONE rather than matching specific types. From what I can see DetermineHdrType never returns HDR10PLUS, the plus only gets discovered from the bitstream and reported back through process info, so HDR10+ content seems to reach the codec still typed as HDR10. DV files whose container lacks the config record (older mkv muxes, TS without the descriptor) also appear to type as HDR10 while still carrying RPUs.

xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAmlogic.cpp
matthane: Got it, done.

matthane:
> Instead "move" please revert complete [Add Amlogic display mode, EOFT and colourimetry](https://github.com/CoreELEC/xbmc/commit/9c86cd00798af12f779e5f463258f92f62110e45) and [alsa: provide ALSA output channel layout](https://github.com/CoreELEC/xbmc/commit/793587be5e09ca832a09218b6635d8dfa2893136) and make a new implemention.
> 
> Otherwise it will continue fail on rebase as it's still be applied to core code and then removed again.
> 
> Once merged I will drop whole commit and revert on a rebase afterwards.

Done. Both commits are fully reverted in their own commits now and the labels come back as new implementations on the registry, same pattern as the kernel version one. From what I could see the ProcessInfo getter and its debug log had no other users, so I left them out of the re-add, but I can put them back if you'd rather keep them. Once you drop the originals on a rebase the reverts can go with them.

xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAmlogic.cpp
```diff
@@ -6,12 +6,15 @@
  *  See LICENSES/README.md for more information.
  */
 
+#define AML_FRAME_METADATA_LATCH
```
Portisch: I think we can skip this compiler flag?
Any reason why compile without this feature?

xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAmlogic.cpp
```diff
@@ -327,6 +343,9 @@ bool CDVDVideoCodecAmlogic::Open(CDVDStreamInfo &hints, CDVDCodecOptions &option
             m_hints.dovi.dv_profile = 8;
             m_hints.dovi.el_present_flag = false;
             m_bitstream->SetConvertDovi(true);
+            if (!m_streamMeta.flags.empty())
+              m_streamMeta.flags += ' ';
+            m_streamMeta.flags += "converted";
```
Portisch: Better use `std::vector<std::string>` or `std::set<std::string> or std::unordered_set<std::string>`.
Then the split by `' '` can be avoided. Same about the others if need:
```c
struct AMLFrameMetadata
{
  std::string doviConfig;
  std::string flags;
  std::string structure;
  std::string doviRpu;
  std::string hdr10pSei;
  std::string hdrMdcv;
  std::string hdrCll;
```

Portisch (commented):
I was struggling right now with the broken screenshot by xbmc/master.
They complete broke video frame capture on vfm systems.
So only GUI capture is possible right now untils it's fixed.

So I continue here as I see the PR looks like to be finished soon.

Please check the new review comments.

xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAmlogic.cpp
matthane: Removed, there is no reason to build without it. The parser section compiles unconditionally now.

xbmc/cores/VideoPlayer/DVDCodecs/Video/DVDVideoCodecAmlogic.cpp
matthane: Done. flags is a std::vector internally and the label publishes it as a JSON array now, so nothing needs to split the value anymore. The other members each hold a single payload so they stayed plain strings. I've updated script.module.sidedata now to read the array form.

matthane:
Both comments are addressed and it's rebased onto the current branch. The revert messages reference the renumbered commits. Verified on the device with a fresh build.

Portisch:
You have also a own repo with a add-on example?

So if the implementation does work correctly and all need features are added before merge.

matthane:
Yes, I have the module in my own repo here: https://github.com/matthane/script.module.sidedata

And a testing addon that uses the fields from it here: https://github.com/matthane/script.dvhdr.labels.diagnostic

If possible, after the merge I could PR the module into CE repo and that way it would be easy for addons or skins to use it as a dependency. The only question mark is the bundled libdovi.so the module uses for rpu parsing. CE's libdovi is built as a static lib for kodi, so there is nothing on the device to open at runtime and the module has its own version bundled. HDR10+ parsing borrows CE's libavutil since that one does ship as a shared library. If a bundled binary in an addon isn't wanted, that would have to be addressed first.

Portisch:
Ok, then I will merge this now for the next nightly.
A "cleanup" will be done on a new rebase after some time when there do not pop up any other item.

Please make a new PR for a example of the script addon:
https://github.com/CoreELEC/CoreELEC/tree/coreelec-22/projects/Amlogic-ce/packages/addons/script

CE-Repo:
I'm so happy. Awesome.
Thanks @Portisch

matthane:
Thanks for the time and feedback and helping get this merged. I'm really glad we were able to add this feature! I'll get the module PR'd next.

PR #69, opened 2026-08-17
VideoPlayer/GLES: HDR PGS subtitles pre-baked to BT.2020 sRGB with GUI 709->2020 shader path

simonhust:
Problem
    
      UHD-BD HDR PGS subtitles are authored in BT.2020/PQ, but the stock FFmpeg
      PGS decoder converts the stream YCbCr palette with a hardcoded BT.709
      matrix. When the renderer then treats the result as plain sRGB, subtitles
      get double transfer/gamut errors (washed out or wrong colors on HDR/DV).
    
Changes
    
      - GUI shaders: the KODI_TRANSFER_PQ branch now applies the linear
        BT.709->BT.2020 gamut matrix in the sRGB-encoded domain; the sRGB->PQ
        transfer is left to the kernel OSD HDR2 core.
      - DVDOverlayCodecFFmpeg: PgsIsPqAuthored() (HDMV PGS + video HDR10/HDR10+/DV)
        marks the stream at decode time; overlays carry m_isHdrPq.
      - VideoPlayer: propagate the video hdrType to the subtitle stream hint so
        the subtitle decoder can decide PQ authorship.
      - OverlayRendererUtil: prebake_hdr_pgs_palette() converts the 256-entry
        palette from PQ BT.2020 (decodePQ -> 709->2020 -> fixed 203-nit peak scale
        -> sRGB encode) once per texture, zero per-pixel cost.
      - OverlayRendererGLES: HDR PGS textures render with the new
        SM_TEXTURE_NOBLEND_NO_PQ variant so the 709->2020 transfer path is not
        applied a second time; SDR overlays keep the regular method.
      - RendererAML: set osd_gamut_bypass=1 for HDR output since the OSD is
        already BT.2020 in userspace.
    
Kernel dependency
    
      Requires the common_drivers change "amvecm: fix OSD color mapping on
      HDR/DV playback" (osd_gamut_bypass + SDR_HDR OSD routing) to be present;
      without it the OSD gamut is converted twice.

https://github.com/CoreELEC/common_drivers/pull/34


simonhust:
Samples.
https://drive.google.com/drive/folders/12WTkMIa0S1ywfbvGs50TgW148o7PY43B

Portisch:
Do become https://github.com/CoreELEC/xbmc/commit/7c2133008e7715cd3e1cb4306d25486227e30621 obsolete with this PR?

simonhust:
ok，i cancel it.

simonhust:
> Do become [7c21330](https://github.com/CoreELEC/xbmc/commit/7c2133008e7715cd3e1cb4306d25486227e30621) obsolete with this PR?

I just reviewed this commit. Its purpose is not entirely the same as my PR's, because the HDR PGS content is in the BT.2020 color space while the GUI uses BT.709. To prevent the kernel from applying a secondary PQ mapping, I pre-baked the GUI to BT.2020 with the sRGB transfer function in the XBMC user space. This makes it consistent with the HDR PGS data after build_rgba is applied. I then pass this directly to the kernel, and after the PQ conversion, it produces the correct HDR PGS mapping result.

Portisch:
But because you remove `m_sdrPeak` the setting in Kodi becomes useless.

Maybe instead remove of the value undefine `KODI_TRANSFER_PQ` and impement a new define?

Like?
```c
#if defined(KODI_TRANSFER_PQ)
  rgb.rgb *= m_sdrPeak;
#elif defined(AML_TRANSFER_PQ)
  // BT.709 -> BT.2020 gamut; transfer (sRGB -> PQ) done by the OSD HDR core.
  const mat3 bt709_to_bt2020 = mat3(0.6274, 0.0691, 0.0164, 0.3293, 0.9195, 0.0880,
                                    0.0433, 0.0114, 0.8956);
  rgb.rgb = bt709_to_bt2020 * rgb.rgb;
#endif
```

The `m_sdrPeak` is only applied in HDR mode.

CE-Repo:
Could this work for CE as well? The color issue seems to be identical to the one on Android. 

https://github.com/xbmc/xbmc/pull/28996
https://github.com/xbmc/xbmc/issues/28603

Portisch:
Maybe yes, it must just be picked for AML platform.
There are multiple items same compatible like on platform Android/GBM/GLES.

And Android do same as AML, convert the GUI to BT.2020 in kernel, not in user space.
This is why all this HDR GUI rework just don't work with Android or AML, it applies the conversation double.

But it seems the GUI process is missed for DV:
https://github.com/CoreELEC/common_drivers/pull/34/changes/cfaca615898fa71e950680c69bae331565f54f86

bluesea168:
The method is clearly the standard, and it's converted according to the white paper, so why not do it?
On Android Kodi, it's because the subtitles are bundled with the GUI, there's nothing you can do, you can only map them together with the GUI.

simonhust:
> Could this work for CE as well? The color issue seems to be identical to the one on Android.
> 
> [xbmc/xbmc#28996](https://github.com/xbmc/xbmc/pull/28996) [xbmc/xbmc#28603](https://github.com/xbmc/xbmc/issues/28603)

Because CoreELEC uses the AML windowing system, many older SoCs—such as G12A/B, SM1, and SC2—do not support multi-OSD HDR functionality. To properly handle the unified mapping of both the GUI and subtitles, compositing must be performed in GLES, and then correct mapping is applied through kernel capabilities. Since the GUI uses BT.709 with sRGB, while subtitles are converted to BT.2020 with sRGB, there are two possible paths: either map the GUI from BT.709 to BT.2020, or map the subtitles from BT.2020 to BT.709. To preserve subtitle accuracy and minimise conversions, I chose to map the GUI.

In fact, correct mapping of HDR PGS subtitles can also be achieved through GLES computations, but this is computationally expensive. Using kernel capabilities significantly reduces the computational cost.

Additionally, because DV (Dolby Vision) OSD processing is a black box for us, whereas HDR2 is transparent, and to support future brightness and saturation adjustments（just like T5M SOC）, it is a reasonable choice to route DV to HDR2 for processing.

simonhust:
> But because you remove `m_sdrPeak` the setting in Kodi becomes useless.
> 
> Maybe instead remove of the value undefine `KODI_TRANSFER_PQ` and impement a new define?
> 
> Like?
> 
> ```c
> #if defined(KODI_TRANSFER_PQ)
>   rgb.rgb *= m_sdrPeak;
> #elif defined(AML_TRANSFER_PQ)
>   // BT.709 -> BT.2020 gamut; transfer (sRGB -> PQ) done by the OSD HDR core.
>   const mat3 bt709_to_bt2020 = mat3(0.6274, 0.0691, 0.0164, 0.3293, 0.9195, 0.0880,
>                                     0.0433, 0.0114, 0.8956);
>   rgb.rgb = bt709_to_bt2020 * rgb.rgb;
> #endif
> ```
> 
> The `m_sdrPeak` is only applied in HDR mode.

Fixed it.`m_sdrPeak` is only bypassed for `HDR PGS` content.

simonhust:
> The method is clearly the standard, and it's converted according to the white paper, so why not do it? On Android Kodi, it's because the subtitles are bundled with the GUI, there's nothing you can do, you can only map them together with the GUI.

https://github.com/CoreELEC/xbmc/pull/69#issuecomment-5322068680

simonhust:
> But because you remove `m_sdrPeak` the setting in Kodi becomes useless.
> 
> Maybe instead remove of the value undefine `KODI_TRANSFER_PQ` and impement a new define?
> 
> Like?
> 
> ```c
> #if defined(KODI_TRANSFER_PQ)
>   rgb.rgb *= m_sdrPeak;
> #elif defined(AML_TRANSFER_PQ)
>   // BT.709 -> BT.2020 gamut; transfer (sRGB -> PQ) done by the OSD HDR core.
>   const mat3 bt709_to_bt2020 = mat3(0.6274, 0.0691, 0.0164, 0.3293, 0.9195, 0.0880,
>                                     0.0433, 0.0114, 0.8956);
>   rgb.rgb = bt709_to_bt2020 * rgb.rgb;
> #endif
> ```
> 
> The `m_sdrPeak` is only applied in HDR mode.

Done.

bluesea168:
CE is not Android Kodi, I disagree with your method

This is the method for picking colors for HDR_Authored subtitles
Import the HDR video into subtitle software, use the eyedropper tool to pick the color from the hardcoded subtitles, and this color is what you need for the soft subtitles. 
Aegisub_selectColor
pgsMapping

Then use ASS-to-SUP subtitle software(for example AssRocket or easyavs2bdnxml) to convert the ASS subtitles to SUP / BDN XML，then import the BDN XML (with the PNG image files) into Scenarist UHD,and Scenarist also outputs SUP(Re-authoring UHD BDMV/ISO with Scenarist UHD).
Hardcoded subtitles are in the HDR video, and UHD Blu-ray player can passthrough HDR or DV. 
Soft subtitles, can't be passedthrough  by the player, the player needs to convert PGS subtitles from 709 to BT2020 PQ according to the Blu-ray Disc Association (BDA) standard,this way, the color of the soft subtitles will match the hardcoded subtitles.
cpm and avdvplus convert subtitles according to the BDA standard. 
Can you confirm that your method also follows the BDA standard?

simonhust:
> CE is not Android Kodi, I disagree with your method
> 
> This is the method for picking colors for HDR_Authored subtitles Import the HDR video into subtitle software, use the eyedropper tool to pick the color from the hardcoded subtitles, and this color is what you need for the soft subtitles. Aegisub_selectColor pgsMapping
> 
> Then use ASS-to-SUP subtitle software(for example AssRocket or easyavs2bdnxml) to convert the ASS subtitles to SUP / BDN XML，then import the BDN XML (with the PNG image files) into Scenarist UHD,and Scenarist also outputs SUP(Re-authoring UHD BDMV/ISO with Scenarist UHD). Hardcoded subtitles are in the HDR video, and UHD Blu-ray player can passthrough HDR or DV. Soft subtitles, can't be passedthrough by the player, the player needs to convert PGS subtitles from 709 to BT2020 PQ according to the Blu-ray Disc Association (BDA) standard,this way, the color of the soft subtitles will match the hardcoded subtitles. cpm and avdvplus convert subtitles according to the BDA standard. Can you confirm that your method also follows the BDA standard?

Because my previous modifications were rather messy, I will try to compile a test firmware to see the effect. Of course, this is based on my own understanding of HDR PGS subtitle mapping. I cannot guarantee that it complies with BDA or any other standards. If you think there is a better solution, why not port it to CoreELEC yourself？

bluesea168:
I'm not a programmer, so I don't know how to port code.
If there's a better way to do it on CE, why not use it? The Android approach is a workaround born out of necessity.

simonhust:
> I'm not a programmer, so I don't know how to port code. If there's a better way to do it on CE, why not use it? The Android approach is a workaround born out of necessity.

If you run this code through an AI-based code analyzer, you will see that my implementation is functionally equivalent to CPM. However, the `EOTF` conversion in CPM relies on heavy arithmetic operations like `pow()`, which are still non‑trivial for lower‑end SoCs such as the S905X2 and S905X4. Instead of replicating those computations in the shader, I reuse the existing HDR2 processing path inside the kernel, thereby offloading the most expensive calculations to the underlying hardware/driver layer.

https://github.com/simonhust/common_drivers/blob/cfaca615898fa71e950680c69bae331565f54f86/drivers/media/enhancement/amvecm/set_hdr2_v0.c

xbmc/cores/VideoPlayer/VideoRenderers/HwDecRender/RendererAML.cpp
```diff
@@ -77,6 +79,11 @@ bool CRendererAML::Configure(const VideoPicture &picture, float fps, unsigned in
 
   CServiceBroker::GetWinSystem()->GetGfxContext().SetTransferPQ(dv_is_used | hdr_is_used);
```
Portisch: Is this then still need? Or better complete or partial revert https://github.com/CoreELEC/xbmc/commit/7c2133008e7715cd3e1cb4306d25486227e30621 before?

bluesea168:
Kimi AI
PR #69 CPM subtitle color accuracy analysis

CPM is more color-accurate.
Here's why:
1. Specification Compliance
CPM strictly follows the BD-ROM Version 3.1 White Paper subtitle compositing specification: PGS subtitles retain their native BT.2020 ST.2084 format in HDR mode, with 8-bit values scaled ×4 to 10-bit before direct compositing with the video. This is the rendering path intended by Blu-ray authors.
PR #69, by contrast, downgrades subtitles to sRGB 8-bit. While it fixes FFmpeg's hardcoded BT.709 matrix error, sRGB's gamut and dynamic range are dramatically smaller than BT.2020 PQ — it is fundamentally a lossy conversion.
2. Luminance Fidelity
 
PR #69: Locks 203-nit reference white to sRGB 1.0, then clamps. Any HDR subtitle highlights above 203-nit are hard-clipped and lost. sRGB peak brightness is only ~100 nits, which does not match HDR displays capable of 1000+ nits.
 
CPM: Preserves the full PQ curve. Subtitle luminance relationships are handled correctly by the display's tone mapper, matching the creator's intent.
3. Conversion Chain
 
PR #69: PGS (PQ BT.2020) → linear light → sRGB → texture → Amlogic kernel converts back to PQ/BT.2020 → output. Multiple conversions introduce cumulative errors.
 
CPM: PGS (PQ BT.2020) → CLUT → 10-bit compositing → output. Shorter chain, fewer conversions.
4. Dolby Vision
CPM support Dolby Vision . PR #69 does not support Dolby Vision at all.
5. User Calibration
CPM provides brightness, contrast, and saturation adjustments, allowing users to calibrate for their specific display characteristics. PR #69 uses fixed parameters with no compensation for display variation.
Conclusion: PR #69 is a performance-first engineering fix (making HDR subtitles "viewable"), while CPM is a specification-first accurate implementation (displaying HDR subtitles "correctly"). If color accuracy is the priority, CPM is the superior approach.

Portisch:
What about this, accepted implementation?
https://github.com/xbmc/xbmc/pull/28012

I did try to implement it int past until I realized the BT.709 to BT.2020 is done in hardware.

If all the GUI conversation should be handled in Kodi, hard disable it kernel and use this implementation.
Will it fix also PGS?

This was last status of the implementation for AML platform:
https://github.com/CoreELEC/xbmc/commits/aml-5.15.196-22.0-gui-hdr/

It was before merge of the PR so only a few items are need I guess.
Mostly only this: https://github.com/CoreELEC/xbmc/commit/d7b28c9bdcfa8a0aa65321853e9272d1ab599017.
When I get time again I can try it out.

bluesea168:
You're CoreELEC, you have all the source code. Why do you always tie subtitles and the GUI together? Can't you handle subtitles and the GUI separately?

Portisch:
Because AML video uses VFM, not DRM video plane.
GUI use the OSD system.

There does a hardware decoder for subtitles exist, unused yet.
As ffmpeg does a good job it's not worth to implement a hardware subtitle decoder.
Now are only software decoder supported in Kodi.

bluesea168:
https://github.com/xbmc/xbmc/pull/29054

https://github.com/xbmc/xbmc/pull/29056

Edit: just don't know how Kodi implements the 4× scaling from 8-bit to 10-bit PQ for PGS subtitle brightness.

PR #72, opened 2026-08-24
DVDVideoCodecAmlogic: fix pink Dolby Vision LL playback (regression in nightly 20260816)

cinema-ONE:
Since nightly 20260816, Dolby Vision low-latency playback shows a uniformly pink picture (SDR and HDR10 are fine). Bisected to the 20260815 → 20260816 Kodi bump; introduced by ea1c67bc85 ("DVDVideoCodecAmlogic: latch raw DV/HDR side data from packets").

**Cause:** once frame metadata is being sequenced, `RenderDisplayLatency()` calls `GetOSHDRStatus()` on every `AddData()`/`GetPicture()`, which reads `/sys/class/amhdmitx/amhdmitx0/hdr_cap` and `dv_cap` each time. A read of those attributes can block on the hdmitx driver while a display mode switch is in progress, so at playback start the decode thread stalls for the duration of the refresh rate switch. No frame has reached the DV core when the new mode's colour attributes are decided: `meson_hdmitx_decide_color_attr` logs `[2160p24hz,eotf:0]=>attr[YUV444,10bit]` instead of `forced to YUV422 by Dolby Vision tunneling ... [2160p24hz,eotf:4]=>attr[YUV422,12bit]`, and the DWC video packetizer stays at the GUI packing (`hdmitx_pkt_dump`: VP_CONF selector `pixel_packing`, GCP 30-bit, YCC 422 enable 0) while the AVI/VSIF signal LLDV 422/12-bit — the sink then decodes the stream as scrambled IPT, i.e. pink.

**Fix:** query the OS HDR status at most once per second instead of on every call. The side data feature is untouched.

**Tested:** on two Homatics R 4K Plus boxes, against an LL-only sink and a std+LL sink, all on the 20260816 base for a controlled comparison:
- stock: packetizer left at `pixel_packing`/30-bit, picture pink
- ea1c67bc85 reverted: packetizer `YCC 422`/24-bit, picture correct
- ea1c67bc85 kept + this change: packetizer `YCC 422`/24-bit, picture correct (visually verified)

A secondary observation that supports the analysis: with no display attached the mode switch is effectively instant and the race is never lost, so the bug does not reproduce headless.


Portisch:
I think this speed break stop the purpose of the metadata grabber of each frame.

Instead I would suggest to improve the function `IsHDRDisplay` to use like static values or make a backup in a private class variable so the sysfs is tested only once. The HDR capabilities of a screen will not change from one frame to another.

Portisch:
@cinema-ONE does thi method solve this issue?

If yes I will make a CE coding style change to fix this item.

cinema-ONE:
> @cinema-ONE does thi method solve this issue?
> 
> If yes I will make a CE coding style change to fix this item.

Sorry for the delay, I wanted to test the fix on the hardware. I have now tested it with player-led DV mode. With this change the packetizer switches to YCC-422 as it should and DV-LL playback is visually correct again, tested on the current branch tip on a Homatics R 4K Plus. Please restyle as you see fit. I'll tell Claude Code to have a look, maybe learn for next time.

Portisch:
Fixed by rework of https://github.com/CoreELEC/xbmc/commit/0b42aa8cfd27f9ec5d10e11764b7a13a0e441dea

PR #73, opened 2026-08-25
WinSystemAmlogic: read HDR capabilities on display events

matthane:
Two fixes in the Amlogic window system, following up on the recent hotplug rework (tip a730b2d675b5117d097b9c08740b4605f5f04704).

The main change moves the HDR capability cache onto the event plumbing that rework introduced. With display changes now arriving as hotplug events, hdr_cap and dv_cap can be read exactly where the display changes: once at window system init and again on each hotplug, rebuilding the object every time so it always describes the current display. Everything else, including the per-frame IsHDRDisplay path, becomes a plain read of that cache.

Before the rework there was no reliable in-process signal for a sink change, so lazy caching was the only workable shape. This new approach keeps what that caching fixed in #72 (these sysfs reads can stall on the hdmitx driver during a mode switch) while closing its gaps: the flag-only setters could carry a previous panel's capabilities across a sink swap, the codec could query the cache before anything filled it on a boot straight into playback, and the init flag itself was never initialized. An unplug keeps the last known capabilities on purpose, so a handshake blip cannot latch empty capabilities into a running stream.

The second commit restores an else lost in an earlier rework, so the missing preferred mode warning no longer fires on every hotplug.

Tested on an Ugoos AM9 Pro with an LG C4. The System Info HDR row populates at boot through the new path, and an HDMI replug repopulates it and no longer logs the spurious warning.

### DIFF
--- a/xbmc/windowing/amlogic/WinSystemAmlogic.cpp
+++ b/xbmc/windowing/amlogic/WinSystemAmlogic.cpp
@@ -213,6 +213,7 @@ void CWinSystemAmlogic::HotplugEvent()
     CLog::Log(LOGDEBUG, "CWinSystemAmlogic - HotplugEvent, preferred mode: {}, display mode: {}",
       preferred_mode, CDisplaySettings::GetInstance().GetResolutionInfo(res).strId);
   }
+  else
     CLog::Log(LOGWARNING, "CWinSystemAmlogic - HotplugEvent, no preferred mode defined, use display mode: {}",
       CDisplaySettings::GetInstance().GetResolutionInfo(res).strId);
 
--- a/xbmc/windowing/amlogic/WinSystemAmlogic.cpp
+++ b/xbmc/windowing/amlogic/WinSystemAmlogic.cpp
@@ -188,6 +188,9 @@ void CWinSystemAmlogic::HotplugEvent()
       mode_count);
     return;
   }
+
+  UpdateHDRCapabilities();
+
   std::string preferred_mode = m_amlDisplay->aml_get_preferred_mode();
   RESOLUTION res = static_cast<RESOLUTION>(RES_DESKTOP);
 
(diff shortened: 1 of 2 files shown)


xbmc/windowing/amlogic/WinSystemAmlogic.cpp
```diff
@@ -464,40 +468,41 @@ void CWinSystemAmlogic::UpdateResolutions()
...
 
-bool CWinSystemAmlogic::IsHDRDisplay()
+void CWinSystemAmlogic::UpdateHDRCapabilities()
 {
-  if (!m_hdr_caps_initialized)
-  {
-    CSysfsPath hdr_cap{"/sys/class/amhdmitx/amhdmitx0/hdr_cap"};
-    CSysfsPath dv_cap{"/sys/class/amhdmitx/amhdmitx0/dv_cap"};
-    std::string valstr;
+  CHDRCapabilities caps;
```
Portisch, line 475: Instead a new local variable just reset the class member?
`m_hdr_caps = CHDRCapabilities();`

Portisch (commented):
I was thinking about a "refresh" of the capabilities, but I discarded this idea.
Who has two different displays placed together?

Anyway, please check this `m_hdr_caps` item and we are good to merge.

xbmc/windowing/amlogic/WinSystemAmlogic.cpp
matthane: The local is intentional. Resetting the member first would leave the caps empty while the sysfs reads run, and a stream opening during a hotplug could catch that window and strip DV or HDR10+. Filling a local and assigning at the end avoids it. Added a short comment there so the shape explains itself.

matthane:
I was thinking about AVRs and different standby or power on orders potentially causing issues. Not a guarantee it ever happens, but it seemed worth guarding against since the events are already there.

PR #74, opened 2026-08-26
WinSystemAmlogic: fixes on top of the display code rework

matthane:
This follows the display code rework that moved the display state into AMLDisplay and the capability reads onto DRM properties. The commits restore the behavior fixes from the earlier revision of this PR that did not survive the migration, plus four corrections found while rebasing onto the new code.

The main themes are that the capability refresh runs before the ready gate and no longer writes stored setting values, so a disconnect, an EDID swap, or a boot with no EDID on the wire cannot freeze the verdict rows or wipe user settings. The remaining commits are single purpose fixes and carry their own reasoning.

Tested on an S905X5-J box through an AVR that can strip DV and HDR from the EDID live. Covered a boot with no EDID present, EDID swaps in both directions with the settings page closed, and with it open, where the rows come out right after backing out and re-entering, plus a cold replug with the AVR waking the link afterwards. DV engaged again in the same Kodi session once the sink returned, no restart needed. Stopping a dolby vision playback now leaves the GUI clean, where before the DV core stayed latched and discolored it.


xbmc/windowing/amlogic/WinSystemAmlogic.cpp
```diff
@@ -491,15 +492,22 @@ void CWinSystemAmlogic::UpdateHDRCapabilities()
       caps.SetHLG();
   }
 
+  // missing or empty dv_cap counts as supported
+  bool dv_supported = true;
```
Portisch: Now you default to DV support even when dv_cap path does not exist?

xbmc/windowing/amlogic/WinSystemAmlogic.cpp
```diff
@@ -491,15 +492,22 @@ void CWinSystemAmlogic::UpdateHDRCapabilities()
...
+
   if (dv_cap.Exists())
   {
     valstr = dv_cap.Get<std::string>().value();
     if (valstr.find("2160p30hz: 1") != std::string::npos)
       caps.SetDolbyVision();
     else if (valstr.find("2160p60hz: 1") != std::string::npos)
       caps.SetDolbyVision4k60();
+
+    dv_supported = valstr.find("The Rx don't support DolbyVision") == std::string::npos;
```
Portisch: Use directly `dv_supported = caps.SupportsDolbyVision() != DOLBYVISION_TYPE_NONE`?

xbmc/windowing/amlogic/WinSystemAmlogic.cpp
matthane: That is the behavior carried over from before. The old aml_display_support_dv() left its static at -1 when the node was missing, and -1 returned as bool is true, so a missing dv_cap already counted as supported. I kept it so the only change here is the verdict staying live.

xbmc/windowing/amlogic/WinSystemAmlogic.cpp
matthane: From what I can see they would come out the same on the 5.15 kernel, since a block that parses correctly always prints one of the 2160p lines. But that line reads more like a formatter convention than the verdict, for V2 blocks the kernel prints 2160p30 as a floor even when the sink may not do 2160p at all, so I kept the deny line since it is the kernel's own verdict and it keeps this the exact old check made live. If you prefer the caps check I can change it.

xbmc/windowing/amlogic/WinSystemAmlogic.cpp
Portisch: Then this was a bug, please fix it for the rework.

xbmc/windowing/amlogic/WinSystemAmlogic.cpp
Portisch: Yes, it should match kernel. I think 2160p30  is set as default anyway if DV support. It's only about the minimum 597MHz clock need for 60hz.

xbmc/windowing/amlogic/WinSystemAmlogic.cpp
matthane: Fixed in the rework. The verdict now derives from the parsed capabilities, so a missing or unreadable dv_cap counts as unsupported, and the flag starts out false until the first real read.

xbmc/windowing/amlogic/WinSystemAmlogic.cpp
matthane: Changed to the caps check. The clock detail matches what I found in the kernel source, so the deny line is gone entirely.

Portisch:
Sorry, this PR just don't work as it should. I boot with DV display, disconnect HDMI. Changed TV (EDID) to a non DV TV and kodi shows this on HDMI reconnect:
screenshot00001

I also think to update `UpdateHDRCapabilities` before from using sysfs to use DRM property.

matthane:
The capability refresh sat below the not-ready early return, so a disconnect or a not-ready reconnect skipped it with no retry and the verdict stayed frozen on the previous sink. The new commit moves the refresh above that gate.

Here that fixes it, pulling the cable now drops the verdict and hides the sink settings right away, and a reconnect brings them back with DV playback working without a restart. ~I could only test with cable pulls and an AVR power cycle though~ I found how to disable DV on the AVR and also test that way and it's working for me now, but could you check if it also covers your EDID swap? If you're on the settings page when you swap EDID, you'll have to back out and back in to see them change, but there should be no bugged state anymore.

On the DRM property, from what I can see it reads the same live kernel state as the sysfs node, so the ordering comes out the same. I can move the reads over in a follow up if you prefer it for consistency.

matthane:
We should consider the new commit 26c571ea30c809e462853d3f49484751187b0142 as a bug fix/gap closure to #73 rather than part of this series. The not-ready return also skipped the capability refresh, so a disconnect could leave the HDR capabilities frozen even before these changes.

It stands alone at the tip, so it could be cherry-picked directly or folded in on a rebase, and I can drop it from this PR in that case.

Portisch:
I moved now all display releated code from AMLUtils to AMLDisplay and I picked your last commit about visibility of the settings. Now the settings hide and seek as it should.

Please check it so we can close this PR.

matthane:
Awesome, thanks! I'm reviewing now and will test on device too. I'll report back and/or close the PR soon once I can verify as well.

matthane:
The move into AMLDisplay and the DRM property read are a clear improvement over what I had, thanks for the rework. I rebased this PR onto your branch and repurposed it for the pieces that did not make it across the migration, plus a few things I hit while rebasing onto the new code.

The short version is that the capability refresh now runs before the ready gate and never writes stored setting values, and the commits carry the rest one fix at a time.

Retested on my box with live EDID swaps both directions through the AVR, a dark boot, and a cold replug. That includes the settings page case you raised, where the rows land correctly after a back-out and re-enter. DV engaged again in the same session once the sink returned, no restart needed.

xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
```diff
@@ -2462,6 +2462,20 @@ void CAMLCodec::CloseDecoder()
...
+
+    // the transition needs a few display frames before the core powers down
+    CSysfsPath dolby_vision_status{"/sys/module/aml_media/parameters/dolby_vision_status"};
+    if (dolby_vision_status.Exists())
+    {
+      std::chrono::time_point<std::chrono::system_clock> now(std::chrono::system_clock::now());
+      while (dolby_vision_status.Get<int>().value() != 0 &&
+             (std::chrono::system_clock::now() - now) < std::chrono::milliseconds(250))
+        usleep(10000); // wait 10ms
+    }
```
matthane, line 2477: The wait here gives the driver the few vsyncs it needs with enable still set before it powers the core down, so pulling enable right away can leave the core latched in its last mode. If you would rather handle that in the kernel so the disable completes the teardown on its own, this block can shrink to just the mode write.

Portisch:
I go to squash most into existing commits on next rebase.
Please check after update of the branch.

xbmc/cores/VideoPlayer/DVDCodecs/Video/AMLCodec.cpp
Portisch: `dv_mode` is reset by kernel itself because of `dolby_vision_policy`. But I will keep the `dolby_vision_status` when DV was used.

matthane:
Tested the updated branch on my box. DV profile 5 and profile 7 FEL, both TV-led and player-led, several start/stop cycles each, and the teardown came back clean every time with no leftover DV state or purple screen.

One small thing I noticed in the squash. The settle poll now uses std::chrono::milliseconds(m_decoder_timeout), but that value is in seconds like in the dv_video_on wait just below it, so the cap works out to around 5ms. The comment above the loop still expects a few display frames of wait, but with the 10ms sleep the loop can only run once before the cap expires, so the code doesn't quite do what the comment says anymore. std::chrono::seconds(m_decoder_timeout) or the fixed 250ms would line the two back up.

Thanks for carrying the fixes across, closing this.

Portisch:
Looks like I missed the millisecond item, will fix before nightly build.

PR #76, opened 2026-08-28
Bitmap subtitle zoom and position

cinema-ONE:
Rebased onto the current tip now that the NAL length size fix has landed; this is a single commit against it.

## Description

Picture-based subtitles (PGS, VobSub) are drawn at the size and position the disc authored, and Kodi's "Position on screen" and "Vertical margin" settings do not apply to them. On a frame-packed 3D projector the authored size is often too large to read comfortably, and there is no way to move them off the bottom edge.

This adds two settings:

- **`subtitles.bitmapzoom`** — resize picture-based subtitles, 100% being the authored size
- **`subtitles.bitmapposition`** — apply the existing position and vertical margin settings to them

Only bitmap *subtitles* are affected. DVB subtitles are deliberately excluded: the same codec path also delivers broadcast logos and graphics, which belong where the broadcaster put them. Menu graphics (Blu-ray IG/PG, DVD button highlights) share the overlay classes and are never moved.

## Provenance

This is a carry of **[xbmc/xbmc#29060](https://github.com/xbmc/xbmc/pull/29060)**, which is open upstream, reviewed by reardonia and milestoned `Q* 23.0 Alpha 1` — so it is queued for Kodi v23 rather than v22. I am offering it here early because the feature is most useful on exactly the frame-packed 3D hardware CoreELEC serves, and v23 is months out. If upstream merges it, this commit disappears in your next rebase.

String ids are **39213-39218**: 39211/39212 are taken by CoreELEC's own "Don't show subtitles matching the audio language" setting.

## Testing

Homatics Box R 4K Plus, frame-packed 3D over HDMI to a JVC projector. Zoom at 60% and position enabled, verified on screen with PGS subtitles. Defaults (100%, position off) leave rendering byte-identical to before — checked against an unpatched build on a real PGS clip.

---


cinema-ONE:
Closing this myself - it is not worth your maintenance.

This is an early carry of xbmc/xbmc#29060, which is milestoned for Kodi v23, so CoreELEC
would be carrying +556 lines across every rebase for something that arrives upstream on
its own. That is exactly the cost you raised on #77, and I think you were right about it.

The branch stays at cinema-ONE/xbmc-coreelec:bitmap-subtitle-zoom-position if it is ever
wanted before v23 lands. Nothing here is abandoned - it is just in the wrong place.

PR #78, opened 2026-08-28
AMLDisplay: decode display Dolby Vision VSVDB

matthane:
Kodi already reads dv_cap to tell whether the display supports Dolby Vision, but that only covers the mode bits; the numeric panel data in the block, like peak brightness, is only printed as raw hex and goes unparsed. This adds a small decoder for the block, validated against edid-decode, and makes the result available where it is useful:

- System Info shows the display's peak brightness next to the Dolby Vision entry. It is the only field internal code consumes right now, carried in CHDRCapabilities as an optional that stays unset on other platforms, and more fields can follow the same way once something needs them.
- a new System.vsvdb infolabel returns the full decoded block as one JSON object, same convention as Player.Process(video.sidedata), for skins and addons. Per-field access is planned in the Sidedata addon, which already turns the sidedata label into window properties for skins, so consumers keep one access point and this PR does not grow per-field infolabels
- the block is read once at window system init and refreshed on hotplug, from the dv_cap sysfs node. The DRM property carries only the capability bits, not the block itself, so a sysfs read is unavoidable. But, the handler just formats already-parsed EDID state without taking any lock, so the read cannot stall or contend with mode switches.

This is somewhat groundwork, because I'm thinking of some possible features that adapt DV playback to the connected display which would rely on knowing the peak brightness. A concrete example would be exposing the CMv4.0 append function added in libdovi 3.4.0, where deciding when to append depends on the display's peak brightness. Regardless, it's also valuable debug/nerdy info that would be nice to make more accessible.

Tested on an LG C4, where the decoded 1037 nit peak shows up in the log, in System Info and in System.vsvdb, and refreshes across an HDMI replug. The C4 block is version 2, so the decoder was also run over every distinct VSVDB in the linuxhw/EDID collection of real hardware EDIDs, 333 blocks across versions 0, 1 and 2 from around a hundred display and AVR models, and its output matches edid-decode on every field.

System Info showing the decoded peak brightness


Portisch:
So much code for only one value?

As the EDID can be read by sysfs maybe a external add-on is a better place. Then feed the value to Kodi, when it's need.

matthane:
I know, the self-contained decoder and documentation takes up the majority of the diff and definitely bloats it.

But, I don't see how to get around it because the value would be needed inside Kodi itself where an add-on couldn't reliably deliver it. Peak brightness feeds CHDRCapabilities so that something like libdovi's CMv4.0 append feature could be smartly integrated. The closest an add-on could get is writing a setting over JSON-RPC, but that misses the two moments the value has to be right. At boot the caps are read during window system init before any add-on has started, and on hotplug they refresh straight off the udev event, where no display change notification reaches Python at all, so an add-on would be left polling sysfs and going stale every time the TV or AVR power cycles.

The add-on direction is still useful though for part of this. I'll be able to update the script.module.sidedata add-on to start reading the new System.vsvdb infolabel and provide the individual fields to both add-ons and skins, same as the sidedata label today.

Portisch:
I going as next step to modify everything to become available as DRM property instead sysfs.

This should also include the RAW EDID.
So maybe transfer this to the side data addon should be the better way.

matthane:
Thanks, that works well! A DRM property for the peak brightness would cover the Kodi side directly and would be a cleaner split than carrying a decoder in Kodi. I was originally trying to stay within Kodi space, but going forward I'll consider the kernel when a change makes more sense there. I'll move the full VSVDB decode into the script.module.sidedata add-on instead, so add-ons and skins get the individual fields from the same place they already read sidedata today.

I'll close this now since I think your direction will solve my request.

PR #81, opened 2026-09-01
AMLDisplay: derive the fractional refresh rates in double

allolive:
## Description

The 1000/1001 refresh modes are derived with a `1.001f` divide. That float is not exactly 1.001, so the resulting rate lands one ULP below the rate the player derives from the same ratio in double — as close as two floats can be without being equal, which is enough to defeat an equality test.

This derives the three fractional sites in `AMLDisplay.cpp` in double instead, via a small `FractionalRate()` helper, so the mode rate and the content rate compare bit for bit. Integer modes never divide and are untouched.

## Motivation and context

`CRenderManager::CheckEnableClockSync()` decides whether the display is already running at the content rate with `if (diff && diff < 0.0005)` — the `diff &&` being the perfect-match guard added in 74c3edd2f5 ("RenderManager: CheckEnableClockSync: Do not do a vsync adjust on perfect match").

On Amlogic that guard can never fire for a fractional rate, because `diff` is small than 0. Clock sync therefore stays *enabled* on what is actually a perfect match, `m_vSyncAdjust` becomes non-zero, and `CDVDClock::ErrorAdjust()`
switches from `adjustment = error` to `adjustment = ±m_frameTime` and calls `Discontinuity()` — moving the master clock by a whole frame (41.7 ms at 23.976). The visible result is a picture drawn twice or skipped, worst on panning shots.

Integer rates (24.000, 25.000, …) were never affected: they never divide, so they already compared equal and already had clock sync off.

## How has this been tested?

Amlogic S922X (g12b), Ugoos AM6B Plus, CoreELEC 22 `Amlogic-ce` build.

- Numerically verified the float values either side of the change. After it, 23.976, 29.97 and 59.94 are bit-identical between the mode and the content rate, so the display/content ratio is exactly `1.0` / `2.0`. 3:2 pulldown (23.976 content on a 59.94 display) still yields 2.4999999 and is correctly *not* treated as a match.

Fixes intermittent frame duplication/drops on Amlogic when the display is running at exactly the content rate — i.e. 23.976 / 29.97 / 59.94 content with "Adjust display refresh rate" enabled. Integer frame rates are unaffected.

NA

## Types of change

## Checklist:

### DIFF
--- a/xbmc/windowing/amlogic/AMLDisplay.cpp
+++ b/xbmc/windowing/amlogic/AMLDisplay.cpp
@@ -23,6 +23,16 @@
 #include "utils/RegExp.h"
 #include "windowing/GraphicContext.h"
 
+namespace
+{
+constexpr float FractionalRate(float rate)
+{
+  // Divide in double: in float the result lands one ULP low. Widened explicitly
+  // for -Werror=double-promotion.
+  return static_cast<float>(static_cast<double>(rate) / 1.001);
+}
+} // unnamed namespace
+
 void FbDestroyCallback(gbm_bo* bo, void* data)
 {
   drm_fb* fb = static_cast<drm_fb*>(data);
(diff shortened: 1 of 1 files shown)


allolive:
@Portisch This change seems to have been lost in a force-push just after it was merged. Is it on purpose or a side effect?

Portisch:
Yes, looks like it was gone with https://github.com/CoreELEC/CoreELEC/commit/280bb2d1ac9f688cfdb3cb47e8c5d30086f2cecb.
Sorry, will be fixed with new nightly.

PR #82, opened 2026-09-01
Whitelist: do not match 3D modes for 2D content

cinema-ONE:
### Symptom

With a whitelist holding 2160p entries for 2D and the 1080p24 / 1080p23.976
frame packing entries for 3D MVC, **every 2D 1080p24 file switches the display to
1080p instead of staying at 2160p**. The 3D entries are what a 3D MVC setup needs
in the whitelist, so there is no configuration that gets both.

From my own log, with the two searches next to each other:

```
[WHITELIST] Searching the whitelist for: width: 1920, height: 1080, fps: 23.976, 3D: false:(0x4)
Display resolution ADJUST : 1920x1080 @ 23.98 - Full Screen (53)     <- wrong
[WHITELIST] Searching the whitelist for: width: 1440, height: 1080, fps: 25.000, 3D: false:(0x4)
Display resolution ADJUST : 3840x2160 @ 25.00 - Full Screen (23)     <- right
```

The second one is only right because nothing in the whitelist matches at 25 fps.

### Cause

`FindResolutionFromWhitelist()` accepts a mode when
`(info.dwFlags & dwFlags) == dwFlags`, which asks whether the mode carries *at
least* the wanted flags. For 2D output `dwFlags` is only
`D3DPRESENTFLAG_PROGRESSIVE`. A `disp_cap_3d` entry carries that bit too, so it
satisfies a 2D search, and being an exact width/height/rate match it also takes
the lowest penalty and wins.

### Fix

Two parts, both needed:

1. Require the wanted flags as before, and additionally reject a mode carrying a
   3D flag when none was asked for. Comparing the whole `D3DPRESENTFLAG_MODEMASK`
   for equality, as upstream does, would not work here: one `disp_cap_3d` line can
   set TB and FP together, so `PROGRESSIVE|TB|FP` must still satisfy a
   `PROGRESSIVE|FP` search.

2. Derive the wanted flags from the render stereo mode instead of `is3D`. `is3D`
   is true only when the stream carries a stereo tag, while the switch already
   reads the stereo mode, whose OFF and MONO cases add no flag. Untagged SBS or
   TAB content switched into a stereo mode by hand leaves `is3D` false, and only
   the subset test let it reach the matching 3D mode at all — rejecting 3D modes
   without this would break that case for good. The two searches further down the
   function that are guarded by `stereo_mode` alone already treat it as the
   authority.

### Testing

Built as a CoreELEC 22 image and run on a Homatics Box R 4K Plus driving a JVC
D-ILA projector, whitelist as above.

| case | before | after |
|---|---|---|
| 2D 1080p23.976 | 1920x1080 @ 23.98 | **3840x2160 @ 23.98** |
| 2D 1080p24 | 1920x1080 @ 24.00 | **3840x2160 @ 24.00** |
| 3D MVC, frame packing | 1920x1080 @ 23.98 | 1920x1080 @ 23.98 (unchanged) |
| untagged 2D started with the renderer in TB | 1920x1080 | 1920x1080 (unchanged, now `0x14` rather than by accident) |

The third and fourth rows are the regression checks. The 3D MVC search logs
`3D: true:(0x24)` and still selects the frame packing mode; the fourth logs
`3D: false:(0x14)`, i.e. the TB flag now comes from the stereo mode rather than
from the stream tag.

The flag logic is also covered by a small test that extracts the helper from
`Resolution.cpp` and the flag values from `GraphicContext.h` at build time, so it
cannot drift. Against the old condition exactly one case flips — the reported bug
— which is what says the change is minimal.

### Not addressed here

- `FindResolutionFromOverride()` still compares against `curr.dwFlags` and runs
  before the whitelist, so with `<refreshoverride>` in `advancedsettings.xml` and
  a 3D mode still current, 2D content can pick a 3D mode by that route. Separate
  path, left alone deliberately.
- A user `special://home/userdata/disp_cap_3d` replaces rather than appends the
  mode list, so a file containing only 3D modes would leave a 2D search with no
  candidates. Pre-existing, and the subset test was masking it.


Portisch:
Not sure if this is correct.
I mean `width: 1920, height: 1080, fps: 23.976` should take `1920x1080 @ 23.98`, not `3840x2160 @ 25.00`.

It's a bit confusing as `1080p` is twice in display modes, one time for 2D and a second time for 3D with the flags.
But this means the `1080p` is supported, and if possible also for 3D.
It's the same HDMI/VIC mode for both. The diff for 3D, if supported, is that the 3D mode is additional set by `AMSTREAM_IOC_SET_3D_TYPE`.

The more "correct fix" would be to replace the 1080p entry in the display mode list by the 1080p with 3D options so it apear only once in the list.

cinema-ONE:
Let me state what this whitelist is meant to do, because I think that is where we
are talking past each other.

**Intent: the display stays at 2160p for everything, and only the refresh rate
follows the source. The single exception is stereoscopic playback, which has to drop
to 1920x1080 Frame Packing because that is the only 3D structure the display
accepts.**

In other words the video is scaled to 2160p by the player/SoC, and Kodi is expected
to change the *refresh rate* but never the *output resolution* - except for 3D.

The shape of the list says exactly that:

    0384002160060.00000pstd      2160p60
    0384002160059.94006pstd      2160p59.94
    0384002160050.00000pstd      2160p50
    0384002160030.00000pstd      2160p30
    0384002160029.97003pstd      2160p29.97
    0384002160025.00000pstd      2160p25
    0384002160024.00000pstd      2160p24
    0384002160023.97602pstd      2160p23.98
    0192001080024.00000ptabfrp   1080p24  Frame Packing / Top-Bottom
    0192001080023.97602ptabfrp   1080p23.98 Frame Packing / Top-Bottom

That is 2160p at every refresh rate a source can have, plus 1080p at exactly the two
rates that HDMI frame packing defines. A 2D 1080p24 file is meant to be presented as
2160p24, not to switch the output to 1080p.

Note what is deliberately *not* in the list. This display offers 1080p twice:

    disp_cap      1080p24hz
    disp_cap_3d   1080p24hz FramePacking TopBottom

so there is a `0192001080024.00000pstd` entry available, and it is not whitelisted.
Only the `ptabfrp` one is. The user has said "1080p for 3D only", and the whitelist
can express that because the stored identifier carries the flags.

That is the distinction the subset test erases:
`(info.dwFlags & dwFlags) == dwFlags` with `dwFlags == D3DPRESENTFLAG_PROGRESSIVE` is
true for the `ptabfrp` entry, so a 2D search accepts the 3D mode and, being an exact
size and rate match, it wins on penalty.

Upstream keeps them apart. `FindResolutionFromWhitelist()` there matches with
`(info.dwFlags & D3DPRESENTFLAG_MODEMASK) == (curr.dwFlags & D3DPRESENTFLAG_MODEMASK)`
and uses `is3D` only in the log line, so a 3D entry can never satisfy a 2D search.
23ca3366f3 replaced that with the subset test to add the 3D search - that is where
2D searches started accepting 3D modes. This patch keeps your 3D search and restores
that one property.

---

Also, sorry - the log block in the description was two different cases pasted next to
each other, not one:

    width: 1920, height: 1080, fps: 23.976  ->  1920x1080 @ 23.98   (the problem)
    width: 1440, height: 1080, fps: 25.000  ->  3840x2160 @ 25.00   (correct)

The `3840x2160 @ 25.00` belongs to the 1440x1080 clip. It is only correct because
nothing in the whitelist matches 1080-height at 25 fps.

---

About replacing the two entries with a single 1080p carrying the 3D options: I do not
think that fixes this case on its own. If the merged entry is whitelisted, 2D content
matches it and still drops to 1080p; if it is not whitelisted, the 3D search has no
1080p to find. It would work if 3D selection did not go through the whitelist at all -
which fits what you say about it being the same VIC plus `AMSTREAM_IOC_SET_3D_TYPE` -
but that is a bigger change than this one. Happy to go that way if you prefer it; I
would rather ask than guess.

Measured on a Homatics Box R 4K Plus with a JVC projector:

| case | before | after |
|---|---|---|
| 2D 1080p23.976 | 1920x1080 @ 23.98 | 3840x2160 @ 23.98 |
| 2D 1080p24 | 1920x1080 @ 24.00 | 3840x2160 @ 24.00 |
| 3D MVC frame packing | 1920x1080 @ 23.98 | unchanged, logs `3D: true:(0x24)` |
| untagged 2D, renderer in TB | 1920x1080 | unchanged, logs `3D: false:(0x14)` |

cinema-ONE:
I cannot reopen this myself - GitHub refuses it - so this is a request to reconsider
rather than a reopen. If the argument below does not convince you, please leave it
closed and I will let it go.

The short version: this display offers 1080p twice, once as a 2D mode and once as the
frame packing mode, and the whitelist holds only the second. The stored identifiers are
distinct (`0192001080024.00000pstd` vs `0192001080024.00000ptabfrp`), they are separate
rows in the settings list, and only the `ptabfrp` one is ticked - so 1080p is whitelisted
for 3D and not for 2D. The subset test erases that distinction. Upstream keeps them apart
with an equality test on `D3DPRESENTFLAG_MODEMASK`; 23ca3366f3 replaced it to add the 3D
search, and that is where 2D searches began accepting 3D modes.

I re-checked this branch today: still `MERGEABLE`, and the subset test is still present
12 times on the current tip, so nothing has superseded it.

Portisch:
Closed by squash into https://github.com/CoreELEC/xbmc/commit/66df883a17bd96c70bdd7924a01231452bd3b1cd

PR #83, opened 2026-09-03
AMLCodec: correct the dv_mode values on the DRM property path

matthane:
The sysfs-to-DRM property transition kept the sysfs value offset when moving the dv_mode writes. The DRM property takes the raw output mode, so every write selected the neighboring mode.

Effects with forced conversion active:
- Player-led lands IPT_TUNNEL instead of IPT
- TV-led lands HDR10 instead of IPT_TUNNEL
- In my testing close parks the engine in IPT with the force policy still set, and the stale output attributes break GUI colors until reboot

The three writes now use the raw enum values. Verified on AM9 Pro by reading the mode back at the sysfs node:

| configuration | before | after |
|---|---|---|
| Player-led + sdr2dv | IPT_TUNNEL | IPT |
| TV-led + sdr2dv | HDR10 | IPT_TUNNEL |
| after stop | IPT parked, force policy | off, follow policy |


Portisch:
Thx, I guess this is the reason VS10 is currently broken. Will test and fix it.

matthane:
Yeah I think it is simple in this case. It seems to work as expected with my setup now. I'm also doing a deep dive on the VS10 live mode switches but that is another headache 😅

Portisch:
Closed by https://github.com/CoreELEC/xbmc/commit/a0fb0d6ab6bd160826f3e3ff03def7de37b77956

PR #84, opened 2026-09-06
AMLDisplay: force the modeset when frame packing starts or stops

cinema-ONE:
Frame packing is the same mode name with a doubled timing. `aml_set_drmDevice_active()` only performs
the modeset when `force_mode_switch` is set or the mode name differs from the current one, so when the
display is already at `1080p24hz` - the previous title was 1080p24, or a 3D title in a playlist follows
another - a frame-packed title sends `3dfp` to the driver without a modeset. The driver keeps the 2D
timing: `vinfo` stays at 1080 lines, no VSIF is sent, and a frame-packing projector shows the top eye
stretched with no 3D.

Force the modeset when the stereo mode switches into or out of `HARDWAREBASED`. `m_stereo_mode` is
`UNDEFINED` before the first switch, so the kernel's `stereo_mode` is read for the comparison, as
`handle_display_stereo_mode()` already does.

Measured on an Ugoos AM9 Pro (S905X5M, hdmitx21) against a JVC DLA-RS4100 through an HDFury VRROOM:

- nightly 20260904: playlist `TAB clip (1080p24) -> MVC clip`: `display/mode 1080p24hz`, `stereo_mode 8`,
  `vinfo height 1080`, `3d_info 0`; projector "1080p 24", top half of the frame stretched, no 3D. Same
  clip started from 2160p25: `height 2205`, `3d_info 2`, projector "1080p 24(FP)", correct 3D.
- with this change (same playlist): `height 2205`, `3d_info 2`, and back to `height 1080`, `3d_info 0`
  on stop.

`videoplayer.adjustrefreshrate` = "On start/stop" hides the problem, because every stop returns to
the desktop mode and the next start is a real modeset; "On start" and playlists expose it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/xbmc/windowing/amlogic/AMLDisplay.cpp
+++ b/xbmc/windowing/amlogic/AMLDisplay.cpp
@@ -1037,6 +1037,18 @@ bool CAMLDisplay::set_native_resolution(const RESOLUTION_INFO &res, std::string 
 {
   bool result = false;
 
+  RenderStereoMode previous_stereo_mode = m_stereo_mode;
+  if (previous_stereo_mode == RenderStereoMode::UNDEFINED)
+  {
+    CSysfsPath _kernel_stereo_mode{"/sys/class/amhdmitx/amhdmitx0/stereo_mode"};
+    if (_kernel_stereo_mode.Exists())
+      previous_stereo_mode = static_cast<RenderStereoMode>(_kernel_stereo_mode.Get<int>().value());
+  }
+  if (previous_stereo_mode != stereo_mode &&
+      (previous_stereo_mode == RenderStereoMode::HARDWAREBASED ||
+       stereo_mode == RenderStereoMode::HARDWAREBASED))
+    force_mode_switch = true;
+
   if (aml_get_cpufamily_id() < AML_T7)
   {
     handle_display_stereo_mode(stereo_mode);


Portisch:
The forced mode switch is set here: https://github.com/CoreELEC/xbmc/blob/aml-5.15.196-22.0/xbmc/windowing/amlogic/WinSystemAmlogicGLESContext.cpp#L180

cinema-ONE:
Thanks - that block is the right place to look, and it explains the gap. It only reaches the stereo-mode
comparison (line 179) when the *resolution* is considered unchanged, and that test compares
`iWidth`/`iHeight`. For frame packing the `RESOLUTION_INFO` carries the doubled `iHeight` (2205), so a
switch from a 1080p24 TAB/SBS/2D mode into frame packing is classified as a resolution change, the
branch is skipped, `m_force_mode_switch` stays false, and `aml_set_drmDevice_active()` then sees the
same mode name and does nothing. The driver gets `3dfp` without a modeset and keeps the 1080-line timing.

Reproduced today on the stock nightly 20260905 (`9b78f0024d`) on an Ugoos AM9 Pro, `Adjust display
refresh rate = On start`, video playlist HTAB clip -> MVC clip, debug log:

```
21:14:21.999 CWinSystemAmlogicGLESContext::CreateNewWindow: m_bWindowCreated: true, frac rate 1(1), hdrType: none(none), force mode switch: no
21:14:21.999 CWinSystemAmlogicGLESContext::CreateNewWindow: cur: iWidth: 1920, iHeight: 1080, iScreenWidth: 1920, iScreenHeight: 1080, fRefreshRate: 23.98, dwFlags: 04, nativeGUI: true
21:14:21.999 CWinSystemAmlogicGLESContext::CreateNewWindow: res: iWidth: 1920, iHeight: 2205, iScreenWidth: 1920, iScreenHeight: 1080, fRefreshRate: 23.98, dwFlags: 34, nativeGUI: true
21:14:22.201 CAMLDisplay::handle_display_stereo_mode setting new mode: 3dfp
21:14:22.202 CAMLDisplay::set_display_resolution: try to set mode: 1080p24hz (FramePacking)
21:14:22.202 CAMLDRMUtils::aml_set_drmDevice_active - mode 1080p24hz is already set
```

`/sys/class/display/mode` stays `1080p24hz`, `stereo_mode` 8, vinfo height 1080 (yesterday's
measurement, same sequence); the projector shows the top half of a flat picture. A stop between the two
clips hides it: the stereo-off transition re-runs the whitelist search and the desktop mode comes back,
so the next start is a real modeset.

If you would rather fix it in `CreateNewWindow`, comparing `iScreenWidth`/`iScreenHeight` in the
same-resolution test (and/or forcing when `(current_resolution.dwFlags ^ res.dwFlags) &
D3DPRESENTFLAG_MODEMASK` is non-zero) would reach the existing stereo check. I kept the change in
`AMLDisplay::set_native_resolution()` because that is where the mode name and the 3D mode meet, but I am
happy to rework the PR whichever way you prefer.

*Analysis and text prepared with Claude Code (Anthropic); the measurements were run on the reporter's box.*

PR #85, opened 2026-09-06
Amlogic: hardening fixes for the codec, display and utils paths

allolive:
## Description

A set of independent hardening fixes to the Amlogic paths: unchecked
allocations, pointers dereferenced where they can legitimately be null, resource
leaks, and error returns that lost their sign on the way out.

## Motivation and context

Found by reading the code, not from bug reports, so there is no issue to link.
These are the kind of defect that produces a rare crash or a slow leak rather
than a reproducible failure.

## How has this been tested?

Built for Amlogic-ce (g12b / S922X) against `aml-5.15.196-22.0` and carried in
daily-use builds for several days of playback with no regression.

Each commit applies independently and the series compiles cleanly.

No visible change in normal operation. Fewer rare crashes, and small leaks
removed.

N/A

## Types of change

## Checklist:


Portisch:
Is it ok when I squash most fixes to the releated commits introduce the issues?
This does mean loosing the author of the single commits.

allolive:
Yes you can do that. It's fine.

Portisch:
Squashed by https://github.com/CoreELEC/xbmc/tree/708dda107b5be0f88d485c25f89d6fc2a9e50d17

PR #86, opened 2026-09-09
AML: play HDR10+ as Dolby Vision P8.1 (metadata conversion)

allolive:
## Description

Lets an HDR10+ title play as Dolby Vision on Amlogic devices, using the dynamic metadata the stream already carries.

One new setting, off by default:

    Settings > CoreELEC > Convert HDR10+ to Dolby Vision
    Off (default) / CM v4.0 / CM v2.9

The two modes are versions of the Dolby Vision metadata format. CM v4.0 can carry an average brightness that CM v2.9 has no field for, so it is theoretically a more accurate representation; not every display uses it, which is why both are offered.

The setting is mutually exclusive with the existing VS10 conversion modes, in both directions, and is hidden on devices without Dolby Vision support. Streams that already carry Dolby Vision are left alone.

Amlogic-only: the new files build only under the AML target, and no cross-platform file changes behaviour.

Adapts @cpm-code's HDR10+ to Dolby Vision conversion (from pannal/xbmc, branch aml-4.9-21.3) to Kodi 22, keeping @pannal's switch to libdovi for the RPU and @doppingkoala's histogram-based average.

## Motivation and context

"Tone map HDR to Dolby Vision" does not do this. The VS10 engine already outputs Dolby Vision from an HDR10 stream, but it works from the static values that describe the whole title and never reads the HDR10+ dynamic metadata. An HDR10+ stream is treated as plain HDR10, and the per-scene information it carries is discarded.

This PR makes that metadata usable in Dolby Vision instead of ignored.

## How has this been tested?

Ugoos AM6B Plus (Amlogic S922X-J) running CoreELEC 22.0 Piers, output to an LG G5.

Testing so far covers the mechanism, not the picture. What was checked:

- HDR10+ titles play as Dolby Vision with the setting on; the Dolby Vision path reports the stream as profile 8.1 and stays engaged for the length of playback.
- Both CM v4.0 and CM v2.9 were exercised on the same content, and the two produce the different metadata values they are supposed to.
- Repeated seeking during playback: playback resumes and conversion continues.
- With the setting Off, playback takes the existing path unchanged.
- A stream that already carries Dolby Vision is not converted.

What has NOT been verified is whether the converted result is *colorimetrically accurate*, or how it compares to the same title played as HDR10+ on a display that handles HDR10+  natively.  I have no reference display, no established measurement methodology for that. Nothing here should be read as a claim about picture quality.

Anyone set up to measure this properly would be very welcome to; that is the obvious next step, and the CM v4.0 / CM v2.9 choice in particular deserves it.

On an Amlogic device with a Dolby Vision display, HDR10+ titles can be played as Dolby Vision, driven by the stream's own per-scene metadata instead of the static values describing the whole title. Whether that looks better is not something this PR claims or has measured.

Off by default, so nothing changes unless it is enabled. HDR10+ stored only alongside the video rather than inside it (some remuxes) is not converted and plays as it does today.

N/A — during video playback the Amlogic capture path contains only the video plane, so the result is not meaningfully screenshottable.

## Types of change

- [X] **New feature** (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/addons/resource.language.en_gb/resources/strings.po
+++ b/addons/resource.language.en_gb/resources/strings.po
@@ -8954,6 +8954,20 @@ msgctxt "#14304"
 msgid "RDS Radio"
 msgstr ""
 
+#. Option for setting #14314 - generate a CM v4.0 Dolby Vision RPU
+#: system/settings/settings.xml
+msgctxt "#14305"
+msgid "CM v4.0"
+msgstr ""
+
+#. Option for setting #14314 - generate a CM v2.9 Dolby Vision RPU
+#: system/settings/settings.xml
+msgctxt "#14306"
+msgid "CM v2.9"
+msgstr ""
+
+#empty strings from id 14307 to 14309
+
 #: system/settings/settings.xml
 msgctxt "#14310"
 msgid "Tone map SDR to Dolby Vision"
@@ -8974,7 +8988,19 @@ msgctxt "#14313"
 msgid "Intended for Dolby Vision enabled displays. HDR content will be tone mapped by Dolby Vision VS-Engine."
 msgstr ""
 
-#empty strings from id 14305 to 15010
+#. Setting to convert HDR10+ dynamic metadata to Dolby Vision profile 8.1
+#: system/settings/settings.xml
+msgctxt "#14314"
+msgid "Convert HDR10+ to Dolby Vision"
+msgstr ""
+
+#. Help text of setting with label #14314
+#: system/settings/settings.xml
+msgctxt "#14315"
+msgid "Play HDR10+ video as Dolby Vision. Needs a TV that supports Dolby Vision. Video that already has its own Dolby Vision is left as it is.[CR][CR]CM v4.0 carries an average brightness that CM v2.9 has no field for, so it should be more accurate; not every TV uses it."
+msgstr ""
+
+#empty strings from id 14316 to 15010
 
 #: system/settings/settings.xml
 msgctxt "#14400"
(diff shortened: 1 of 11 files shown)


Portisch:
It just makes no sense to make a mapping from a equivalent format to another.

allolive:
@Portisch I'm of the opposite opinion here, so let me make the case.

Plenty of Dolby Vision TVs don't do HDR10+ at all - LG for one. On those, an HDR10+ title just plays as HDR10 and the per-scene metadata sits in the stream unread/unused. The same logic works the other way round for a Samsung set that does HDR10+ but not DV.

The choice isn't really "HDR10+ or DV", it's "use the scene-by-scene data the file already carries, or throw it away". 

The conversion algorithm in this PR is not an exact translation as the two formats don't define their fields the same way. But they're describing the same concepts, and L1 comes straight from what HDR10+ already carries per frame: maxscl for the peak, the distribution percentiles for the average. The numbers are computed from the available frame statistics.

It's the same kind of feature as the VS10 tone mapping already exposed in the settings. The difference is that VS10 HDR10 to DV mode tone maps from the static HDR10 values. It never reads the per-scene metadata and that data is dropped.

To assess how close the translation lands, I took a few videos that exist in both HDR10+ and Dolby Vision, ran the HDR10+ metadata through this transformation layer, and compared what came out against the file's own Dolby Vision metadata for the same frames. The calculated metadata are not a perfect match, but of similar shape. I wouldn't go and say that the translated image looks accurate as I have no way to verify that, but it should theoretically be more accurate than the base HDR10 image alone.

arthur-liberman:
IMHO, leave HDR10+ to be displayed as HDR10 on such sets. LG TVs can tone map pretty well, and probably better fit the display than transcoding HDR10+ to DV in software on an external device.
Can't say about Samsung and the reverse (DV to HDR10+), but I assume that it can handle HDR10 + tonemapping fine, too.

Portisch:
Enable the HDR to DV option and when display does support only HDR10/DV a HDR10+ media should be "upscaled" to DV automatically.
When the display support HDR10+/DV it uses the matching HDR range to give best results.

PR #91, opened 2026-09-12
AMLCodec: a drain that stalls is still the end of the stream

allolive:
## Description
Playback could get permanently stuck when the next file was started while the current one was finishing. This fixes it.

## Motivation and context
When a file reaches its end, the decoder is asked to hand over whatever frames it still holds. If it cannot hand over any more, nothing ever tells the player to stop waiting, so it waits forever. Starting the next episode at that moment leaves the box frozen on the last picture of the previous one, behind a spinner, using a full CPU core. The only way out is to stop playback by hand.

Now the wait ends on its own after the existing decoder timeout, and playback carries on to the next file.

## How has this been tested?
Ugoos AM6B Plus, CoreELEC 22.0. Play a file to its end and start the next one about a second and a half later: **stuck 4 times out of 4 before the change, 0 times out of 2 after.** Normal playback and normal endings are unaffected.

Choosing "play next" as a file ends no longer freezes playback. It also fixes the wrong, blue-tinted colours that appeared in the menus afterwards.

N/A

## Types of change

## Checklist:


Portisch:
Closed by https://github.com/CoreELEC/xbmc/commit/6c4347c2d2d42656b231b869ed7b9ce1c15b5174

